PR #98, opened 2015-09-23
Remove references to SVN revision IDs, clean up edit summary, minor fixes

fhocutt:


fhocutt:
And split out the classes to individual files.

fhocutt:
re: expandFns and revisionID, I'm leaving those in only in the SQL query-building portion, since nothing is decided about how to use the old database and I don't want the queries to obviously not match up with the recovered schema. I will comment those queries out for now and leave an explicit comment on what revisionID() used to be.

category.php is very outdated and I'm not 100% sure what it does but I'll go ahead and take those out anyway.

fhocutt:
All right, take a look at this--should be fixed up.

fhocutt:
Merged after confirmation from @kaldari of review.

PR #103, opened 2015-12-10
T119932 clean out cite doi

fhocutt:
@kaldari


fhocutt:
@kaldari: took out the references and everything else Template:Cite related.

PR #106, opened 2016-04-30
fixing login issue

kaldari:
updating and hacking Snoopy library, removing code no longer needed in
expandFns.php.

See https://phabricator.wikimedia.org/T128926 for bug report.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1153,7 +1153,8 @@ class Template extends Item {
       $duplicated_parameters = array();
       $duplicate_identical = array();
       foreach ($this->param as $pointer => $par) {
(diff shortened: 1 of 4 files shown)


Template.php
```diff
@@ -1693,23 +1696,7 @@ protected function tidy() {
       else $this->set('others', $others);
     }
 
-    if ($this->added('journal') || $journal && $this->added('issn')) $this->forget('issn');
-
-    if ($journal) {
```
kaldari: Apparently $journal has never been defined, so this never executes.

ms609: Perhaps the line setting it has been removed: $journal would originally have been set to $this->get('journal'), possibly falling back to something like $this->get('periodical') if 'journal' was not set.

PR #109, opened 2016-08-25
Add files via upload

GlazerMann:
Lots of bug fixes

### DIFF
--- a/DOItools.php
+++ b/DOItools.php
@@ -139,6 +139,16 @@ function list_parameters() { // Lists the parameters in order.
      "editor2", "editor2-author", "editor2-first", "editor2-link",
      "editor3", "editor3-author", "editor3-first", "editor3-link",
      "editor4", "editor4-author", "editor4-first", "editor4-link",
+     "editor5", "editor5-author", "editor5-first", "editor5-link",
+     "editor6", "editor6-author", "editor6-first", "editor6-link",
+     "editor7", "editor7-author", "editor7-first", "editor7-link",
+     "editor8", "editor8-author", "editor8-first", "editor8-link",
+     "editor9", "editor9-author", "editor9-first", "editor9-link",
+     "editor10", "editor10-author", "editor10-first", "editor10-link",
+     "editor11", "editor11-author", "editor11-first", "editor11-link",
+     "editor12", "editor12-author", "editor12-first", "editor12-link",
+     "editor13", "editor13-author", "editor13-first", "editor13-link",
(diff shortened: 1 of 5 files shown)


DOItools.php
```diff
@@ -568,7 +578,7 @@ function checkTextForMetas($text){
     $authors = formatAuthors($author_list, true);
     foreach ($authors as $no => $auth) {
       $names = explode (', ', $auth);
-      $newp["author" . ($no + 1)] = $names[0];
```
ms609: Are you sure it's safe to remove this?

ms609:
Looks great, thanks... not quite clear that removing 'author' on line 571 is an unambiguously safe thing to do, but let me know if you think it is.

DOItools.php
GlazerMann: See line below where is it replaced with "last="

GlazerMann: last and author are aliases, so it should be fine.  People prefer last when there is a first, and it is split in this case.

GlazerMann:
Last and author are identical in template, do it is just to make human editors hsppy

ms609:
Makes sense, thanks.  I've merged.

GlazerMann:
We probably need to get this pushed up to the development version for actual use and testing.

Kurt



________________________________

kaldari:
@GlazerMann: This patch breaks Citation bot. Specifically running https://tools.wmflabs.org/citations/doibot.php no longer works.

Template.php
```diff
@@ -405,6 +406,26 @@ public function add_if_new($param, $value) {
...
+      return false;
+      case 'volume':
+        if ($this->blank($param)) {
+          if (  $this->get('journal') == "ZookKeys" ) add_if_new('issue',$value) ; // This journal has no volume
+          return $this->add($param, $value);
+        }
+      return false;
+      case 'bibcode':
+        if ($this->blank($param)) { 
+          $bibcode_pad =  strlen($value) - 19;
```
kaldari: This should be 19 - strlen($value), not strlen($value) - 19.

Template.php
```diff
@@ -405,6 +406,26 @@ public function add_if_new($param, $value) {
...
+          if($bibcode_pad > 0 ) {  // Paranoid, don't want a negative value, if bibcodes get longer
+              value = $value . str_repeat( ".", $bibcode_pad);  // Add back on trailing periods
```
kaldari: Basic syntax error here. Should be $value.

PR #110, opened 2016-08-25
Glazer mann patch 2 - add nowiki support also

GlazerMann:


GlazerMann:
Thoughts on this one?

DOItools.php
```diff
@@ -568,7 +578,7 @@ function checkTextForMetas($text){
     $authors = formatAuthors($author_list, true);
     foreach ($authors as $no => $auth) {
       $names = explode (', ', $auth);
-      $newp["author" . ($no + 1)] = $names[0];
```
ms609: This strikes me as something I might have done intentionally - perhaps I check elsewhere if the author parameter is checked, but not the last parameter.  Have you worked this through for unintended consequences?

Template.php
```diff
@@ -232,16 +232,16 @@ public function add_if_new($param, $value) {
           if (strpos($value, ',')) {
             $au = explode(',', $value);
             $this->add($param, formatSurname($au[0]));
-            return $this->add('first' . (substr($param, -1) == '1' ? '1' : ''), formatForename(trim($au[1])));
+            return $this->add(sanitize_string('first' . (substr($param, -1) == '1' ? '1' : ''), formatForename(trim($au[1]))));
```
ms609: Should sanitize_string surround formatForename(...)?  otherwise there's only one parameter to this->add

Template.php
```diff
@@ -418,23 +439,28 @@ protected function get_identifiers_from_url() {
     if (strpos($url, "jstor.org") !== FALSE) {
       if (strpos($url, "sici")) {
         #Skip.  We can't do anything more with the SICI, unfortunately.
+      elseif (strpos($url, "plants")) {
```
ms609: Can we be confident that the string "plants" would never appear elsewhere in a URL that contains another identifier, e.g. plantscience.com/doi/10.1111/plants.10384?

GlazerMann:
The change here is just swapping author for last.  Note that there is a line added, not just removed

GlazerMann:
I think you counted parenthesis wrong in the first one.  In the second one, I added a new pull request that converts plants to plants.jstor.org.  Good catch.

kaldari:
This is far too many unrelated changes in one patchset. The last time we merged a mega-patch like this, #109, it broke Citation bot entirely and was very difficult to debug due to having so many changes in a single patch. I would prefer that this be resubmitted with unrelated changes in separate patchsets.

GlazerMann:
This commit is messed up.  It is comparing to the old main branch.  Not the new main branch.   This pull needs rebased.   GRRRRRRRRR

GlazerMann:
Rebased in another pull (actually two pulss)

PR #119, opened 2017-09-06
fix ZooKeys better

GlazerMann:
Really need to return before adding volume

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -327,8 +327,7 @@ class Template extends Item {
         return false;
       case "periodical": case "journal":
         if ($this->blank("journal") && $this->blank("periodical") && $this->blank("work")) {
-          if ( sanitize_string($value) == "
-          s" ) $this->blank("volume") ; // No volumes, just issues.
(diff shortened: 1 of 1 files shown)


ms609:
Probably ought to use add_if_new to be safe - in case someone has manually filled the 'issue' parameter with a comment, for example.
Might the code also be clearer if you used the syntax "if (zookeys) add_if_new(issue) **else** add ($param)"?

GlazerMann:
O think it is better now.  Can't believe I misspelled zookeys.

GlazerMann:
One more note.  You might want to delete the branch labeled "wiki".  It has only one file.  Also, the branch labeled "dev" is very out of date.  I assume that it should go too.

GlazerMann:
NOW I am done.  Very sorry about that.

One more note. You might want to delete the branch labeled "wiki". It has only one file. Also, the branch labeled "dev" is very out of date. I assume that it should go too.

ms609:
Sounds good - will the case of ZooKeys always match what you have written or might it ever be "Zookeys"? A case insensitive match would be appropriate in the latter case.  Will merge for now in any case.

PR #121, opened 2017-09-07
read the preg_match man page

GlazerMann:


ms609:
Could you use `stripos` instead of `preg_match` here?  I believe it has a lower computational overhead where you don't need the functionality of a regular expression

ms609:
PS to follow best practice I have created a 'development' branch of the repository, which should be pulled through to the `citations-dev` accoun at wmflabs; I would suggest making and fully testing code changes there first, then – once everything is guaranteed stable – making a pull request from `development` to `master`, so that the updates can be made to the live `citations` account.

PR #122, opened 2017-09-07
I believe this needs to go after UTF-8 stuff

GlazerMann:
It is not doing its job where it is now.

### DIFF
--- a/expandFns.php
+++ b/expandFns.php
@@ -295,7 +295,6 @@ function format_title_text($title, $capitalize = TRUE) {
               : $title
             );
   $title = preg_replace('~[\*]$~', '', $title);
-  $title = sanitize_string($title);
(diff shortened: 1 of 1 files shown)


ms609:
Has the fix been tested?

GlazerMann:
This fix has NOT been tested.  Off course, it probably needs changed in  format_title_text_nocaps() if that is added as per your previous comment.

GlazerMann:
I apologize.  I seem to be targeting master, not development branch.  Now master is ahead of devel :-(

ms609:
Looks to me like the following lines (`$iIn`, `$iOut`) might also belong in sanitize_string?  As it stands, there is some duplication: a pipe `|` will be `&#ed;` before it is `{{!}}`ed.

GlazerMann:
Updated.  You are right.  pipe to {{!}} to escaping is bad.

PR #123, opened 2017-09-07
TitleCase/Sanitize and Unknown as a title

GlazerMann:


ms609:
Are you certain that title-casing titles and journals is always desirable?  Does it correctly handle cases such as "'Studying iPad use in *Homo erectus*', *eLife* 2:222"?

PR #126, opened 2017-09-08
add  format_title_text_nocaps

GlazerMann:


ms609:
Another way of doing this would be to pass a second argument to format_title_text (defaulting to `TRUE`, so that existing calls don't need to be modified) that controls whether words are capitalized.  That way a second function doesn't need to be created and maintained in the future.
i.e. 
```php

function format_title_text($title, $capitalize = TRUE) {
  $title = sanitize_string($title);
  if ($capitalize) $title = capitalize_title($title, TRUE);
// ...
}

format_title_text($title, FALSE);
```

PR #132, opened 2017-09-08
Implement continuous integration with Travis. (Test cases now need implementing.)

ms609:


ms609:
Oops, only want relevant files

PR #137, opened 2017-09-10
no dashes in URLS

GlazerMann:


Parameter.php
```diff
@@ -49,7 +49,8 @@ public function parse_text($text) {
   protected function parse_val($value) {
     switch ($this->param) {
       case 'pages':
-        $this->val = mb_ereg_replace(to_en_dash, en_dash, $value);
+        if ( !preg_match("/http/i", $value)) $value = mb_ereg_replace(to_en_dash, en_dash, $value);
```
ms609: Suggest 
```php
stripos($value, "http") !== FALSE)
```
instead of `preg_match`?

ms609:
Note that the Travis check has picked up a missing bracket.
(I'm working on improving the tests to have a more comprehensive coverage at the moment, so a pass should not be seen as an indication that an edit is problem-free, but a fail is generally worth pursuing.)

PR #142, opened 2017-09-11
copy over some dev changes that worked

GlazerMann:
Formatting title text basically everywhere seems to have fixed the failure to escape evil square braces etc in titles.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -704,7 +704,7 @@ class Template extends Item {
           $this->add_if_new("author$i", $name);
         }
       }
-      $this->add_if_new("title", (string) $xml->entry->title);
+      $this->add_if_new("title", format_title_text((string) $xml->entry->title), FALSE);
       $this->add_if_new("class", (string) $xml->entry->category["term"]);
       $this->add_if_new("author", substr($authors, 2));
       $this->add_if_new("year", substr($xml->entry->published, 0, 4));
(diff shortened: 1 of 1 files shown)


ms609:
Do you have a brief example that would reproduce the bug? Sounds like it would be a useful addition to the suite of test cases I am developing.

ms609:
And is this basically a request to update the master branch from the development branch?  Starting a pull request from the development branch might be a more appropriate way to do that.

GlazerMann:
https://en.wikipedia.org/wiki/User:AManWithNoPlan/sandbox2

Lots of test cases

GlazerMann:
This is basically a request to update master from devel.  Did not do pull from dev, since I am still figuring out GUI.

ms609:
It does take some gettting used to.  At https://github.com/ms609/citation-bot, you want to select Branch:development from the dropdown above the list of files, then click `New pull request` to pull from development into master.  That way the 'edit history' is preserved.

ms609:
I've implemented some of your test cases in the continuous integration suite (see in particular tests/phpunit/TemplateTests.php), so if any future edits cause the errors to recur, they should be flagged right away.
I won't have the time myself, but you might think about incorporating further salient examples from your test suite; let me know if you need a hand figuring out how.

PR #151, opened 2017-09-12
I think this got lost along the way

GlazerMann:


ms609:
Yes - I mean to remark on this - the previous name of the function made it sound (when I looked at it previously) like it would add capitals, but its function is more to remove / standardise them.  Thus I think a better approach is to simply send title_case($string) to the function where this is needed – if nothing else this makes the code easier to read and makes it more explicit what capitalization is being performed and where.  Sorry to give contrary advice when I hadn't reviewed the code in detail!

Perhaps you could double check whether there are any calls to the function where it is necessary to specify an alternative value for `$capitalize`, and we can review whether the parameter is still necessary?

GlazerMann:
I will investigate.

PR #152, opened 2017-09-12
remove tabs also and multiple spaces

GlazerMann:


ms609:
If you want to make the proposed chance you'll need to use `preg_replace`, not `str_replace`.
But I'm not sure the change is necessary; a title may (perhaps) contain multiple consecutive spaces, or tabs, for a reason, and Wikipedia will collapse extra whitespace when it parses the citation.

GlazerMann:
This was suggested on wikipedia talk page as better, but you are right.

PR #159, opened 2017-09-13
don't upgrade dashes if more than one

GlazerMann:
Unless of course they are next to each other.
This avoids changing 1970-10-7 to invalid date 1970–10–7 (really should not be in year, but date).
This also avoids changing 1-4–1-7 to 1–4–1–7 (page 4 to seven of section 1) .
This still changes if it has -- (of course 1-4--1-7 will get hosed still).

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1779,7 +1779,7 @@ class Template extends Item {
             $p->val = preg_replace("~\s+ed(ition)?\.?\s*$~i", "", $p->val);
             break; // Don't want 'Edition ed.'
           case 'year':
-            if (preg_match ('~\d{4}\-\d\d\-\d\d~', $p->val)) { // We have more than one dash, must not be range of years.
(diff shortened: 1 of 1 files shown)


ms609:
What about the case `pp. 5-20; plates 4-6`?
As you say, the year example is a "garbage in, garbage out" example, but if you are concerned about catching it then you could do something like (pseudocode)
```php 
case year:
  if (preg_match ('~\d{4}\-\d\d\-\d\d~', $parameter)) {
   change_parameter_to_date();
    break; 
  }
  // no break here, instead continue through:
case 'pages': case 'page': case 'issue': case 'year':
  if preg_match(PAGES_IN_SECTIONS, $parameter)) {
    preg_replace_appropriate_dashes();
  } else {
    replace_all_dashes();
  }
  break;
```

GlazerMann:
I do not think the preg_match is right yet

GlazerMann:
This pull now detects if a year has this in it:

X-X-X.  Where each X is at least one character.

PR #161, opened 2017-09-19
seems to eat the first item in the template

GlazerMann:


GlazerMann:
This seemed to be what I thought the code needed, but maybe not.

GlazerMann:
Not sure if add or add if new is better idea.

Template.php
```diff
@@ -1371,7 +1371,7 @@ protected function use_unnamed_params() {
       }
       
       if (preg_match('~^(https?://|www\.)\S+~', $dat, $match)) { # Takes priority over more tenative matches
-        $this->set('url', $match[0]);
+        $this->add_if_new('url', $match[0]);
```
ms609: If 'url' is already set, add_if_new should return `FALSE`; in this case, the unused URL might be assigned to a "duplicate_url" parameter?

Template.php
```diff
@@ -1395,13 +1395,8 @@ protected function use_unnamed_params() {
           }
           if ($matched_parameter) {
             $dat = trim(str_replace($oMatch, "", $dat));
-            if ($i == 0) { // Use existing parameter slot in first instance
```
ms609: The objective here is to minimise the change as far as possible, keeping the sequence position of the misspelt parameter, and the spacing as far as possible.

Template.php
```diff
@@ -1395,13 +1395,8 @@ protected function use_unnamed_params() {
...
-              $this->param[$param_key]->param = $matched_parameter;
-              $this->param[$param_key]->val = $match[2][0];
-              $param_recycled = TRUE;
-            } else {
-              $this->add_if_new($matched_parameter, $match[2][$i]);
-            }
+            $this->add_if_new($matched_parameter, $match[2][$i]);
+            $param_recycled = TRUE;
```
ms609: if you use add_if_new, has the parameter been 'recycled'?

Template.php
```diff
@@ -1415,7 +1410,7 @@ protected function use_unnamed_params() {
       }
       if (preg_match("~\(?(1[89]\d\d|20\d\d)[.,;\)]*~", $dat, $match)) { #YYYY
         if ($this->blank('year')) {
-          $this->set('year', $match[1]);
+          $this->add('year', $match[1]);
```
ms609: The difference between `add` and `set` is the reporting printed to the user; not clear that this change is necessary?

Template.php
```diff
@@ -1473,11 +1467,11 @@ protected function use_unnamed_params() {
       ) {
         // remove leading spaces or hyphens (which may have been typoed for an equals)
         if (preg_match("~^[ -+]*(.+)~", substr($dat, strlen($closest)), $match)) {
-          $this->add($closest, $match[1]/* . " [$shortest / $comp = $shortish]"*/);
+          $this->add_if_new($closest, $match[1]/* . " [$shortest / $comp = $shortish]"*/);
```
ms609: Need to catch the case "not new"

Template.php
```diff
@@ -1498,24 +1492,21 @@ protected function use_unnamed_params() {
           case "url":
           if ($this->blank($p1)) {
             unset($pAll[0]);
-            $this->param[$param_key]->param = $p1;
-            $this->param[$param_key]->val = implode(" ", $pAll);
+            $this->add_if_new($p1,implode(" ", $pAll));
```
ms609: Perhaps a check here to evaluate whether `$param_recycled === TRUE` would prevent the same parameter slot being used twice?

ms609:
I can't reproduce this error in a test case - is the issue that the servers are not running the latest versions of the bot?
https://travis-ci.org/ms609/citation-bot/builds/277636169

See 
  public function testMisspeltParameters() in `TemplateTest.php`

GlazerMann:
Might be version is out dated.  Failure is that {{Cite journal|STUFF| pp 3}} becomes {{cite journal|pages=3}} and {Cite journal| pp 3}} becomes {{cite journal}}.

GlazerMann:
Might be version is out dated. Failure is that {{Cite journal|STUFF| pp 3}} becomes {{cite journal|pages=3}} and {{Cite journal| pp 3}} becomes {{cite journal}}.  Can you create a test based upon {{Cite journal| pp 3}}?

I have created two other pulls that incorporate your ideas.

GlazerMann:
It probably is just a messed up version on wiki.

GlazerMann:
I will investigate again if it fails later

PR #165, opened 2017-09-20
param_recycled paranoid

GlazerMann:


GlazerMann:
I do not know if this actually does anything, but this seems like very good defensive programming

PR #167, opened 2017-09-20
add_if_new seems better

GlazerMann:


GlazerMann:
I cannot even find set if null in source tree

ms609:
In which case the calls would cause the bot to fail, meaning either:
- No-one is using this functionality, so we should discontinue it,
or
- We should create a test case that covers this branch of the code, so that similar errors are detected in future.

PR #169, opened 2017-09-20
use last instead of author, if add first too

GlazerMann:


ms609:
What if this is an editor, or if "first1" is already/about to be set and "last1" should be used for consistency?

Perhaps better to

GlazerMann:
This code is only  a change when a full name is passed with authors, author, or authors1 -- not editors.  This means that instead of author1=LAST and first1=FIRST, you get last1=LAST and first1=FIRST.

GlazerMann:
I think you have to see a few lines above the default view to understand this patch.

ms609:
OK, I see how this might make sense.  
How would it deal with non-human authors, e.g. `add_if_new('author', 'BEGS consortium')`?
Perhaps we need to wrap the two lines in an `if (isset($au[1])) { use last and first } else { use $param_name}`?

GlazerMann:
This exact same logic exists for authro2, 3, etc.  Just not author1.

GlazerMann:
I think this is good now. 

When the bot sets author= it try’s to split it into first and last names.  After doing this this change makes last= get set instead of author= for the last name.  This is because we are clearly only setting a last name.  This only applies when splitting author= into two parts.  

The other change is to be willing to set author even if editor is set.

PR #170, opened 2017-09-20
Older PHP does not support const arrays.

GlazerMann:
Probably better to update PHP on Wiki Tools to match test environment.

### DIFF
--- a/constants.php
+++ b/constants.php
@@ -106,24 +106,24 @@ define("common_mistakes", array ( // Common mistakes that aren't picked up by th
 // unCapped is their correct capitalization. Remember to enclose any word in spaces.
 
 
-unCapped = array(' and then ', ' of ',' the ',' and ',' an ',' or ',' nor ',' but ',' is ',' if ',' then ',' else ',' when', 'at ',' from ',' by ',' on ',' off ',' for ',' in ',' over ',' to ',' into ',' with ',' U S A ',' USA ',' et ',
+define("unCapped", array(' and then ', ' of ',' the ',' and ',' an ',' or ',' nor ',' but ',' is ',' if ',' then ',' else ',' when', 'at ',' from ',' by ',' on ',' off ',' for ',' in ',' over ',' to ',' into ',' with ',' U S A ',' USA ',' et ',
 ' ACM SIGPLAN Notices ', ' ASME AES ', ' ASME MTD ', ' BMJ ', ' CBD Ubiquitin ', ' CFSK-DT ', ' e-Neuroforum ', 
 ' Early Modern Japan: an Interdisciplinary Journal ', ' eLife ', ' EMBO J ', ' EMBO J. ', ' EMBO Journal ',
 ' EMBO Rep ', ' EMBO Rep. ', ' EMBO Reports ', ' FASEB J ', ' FASEB J. ', ' FEBS J ', ' FEBS J. ', ' FEBS Journal ',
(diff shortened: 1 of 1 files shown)


GlazerMann:
Seems to me that the test environment might need to match the Wiki PHP version.

ms609:
Oh dear - I'd assumed (hoped) that wmflabs were running PHP7.  The code will require some significant refactoring otherwise - you can't `define` arrays either, so we'd have to go back to using global variables (strongly discouraged).  Can we explore what it would take for wmflabs to upgrade to php7?  It might be as simple as re-naming files to the .php7 suiffix, if they have that version installed

GlazerMann:
It is 5.5.9. 

Old 👵🏻👴🏻

ms609:
That's frustrating. What is the prospect of them upgrading to PHP7?  For now, I've added 5.5.9 to the Travis test cases, but it will take some work to re-adapt the code; probably the thing to do is to make a public function `const_array($name)` that returns the array required, and replace all calls to the constant arrays with a call to the function.  I've no time to do that myself at the moment; feel free to have a shot yourself if that sounds easier than pushing for a PHP version upgrade at WMFlabs!

GlazerMann:
Fighting this issue in wiki land

GlazerMann:
5.6 is supported sort of. Check Wikipedia talk page.

ms609:
I believe that PHP7 is the first version to allow const arrays.

GlazerMann:
Actually 5.6 supports it.  It appears that it was considered useful enough to add.  That’s the last pre-7 version.

ms609:
Oh, great!  Are you able to see what is needed to force the use of this
version?

--

*Martin R. Smith*
Assistant Professor in Palaeontology
Department of Earth Sciences
Durham University
Mountjoy Site, South Road
Durham DH1 3LE

*T*: +44 191 334 2320
*M*: +44 774 353 7510
*E*: martin.smith@durham.ac.uk
*Skype*: martin--smith

durham.ac.uk/earth.sciences/staff/academic/?id=14260
twitter.com/PalaeoSmith

The information in this e-mail and any attachments is confidential. It is
intended solely for the addressee or addressees. If you are not the
intended recipient please delete the message and any attachments and notify
the sender of mis-delivery. Any use or disclosure of the contents of either
is unauthorised and may be unlawful.
Although steps have been taken to ensure that this e-mail and any
attachments are free from any virus, we advise the recipient to ensure they
are indeed virus free. All liability for viruses is excluded to the fullest
extent permitted by law.

GlazerMann:
https://wikitech.wikimedia.org/wiki/Help:Toolforge/Kubernetes#Versions_.26_Packages

Yeah it means using a container.   It’s not just a flag.  This discussion is probably better on Wikipedia

kaldari:
@GlazerMann: Something seems to be messed up with the Kubernetes set up on Toolforge. I haven't been able to get it running at all.

GlazerMann:
The dev code does not work as far as I can tell.  What types of errors do you get.  As a user, we do not see the erros

GlazerMann:
Once we have the error messages; we can fix PHP Code to work with 5.6

ms609:
Even in PHP 5.6, constants must evaluate to scalar values: see
https://travis-ci.org/ms609/citation-bot/jobs/281183252

Looks like the code will have to be refactored.  This might be something I can have a bash at today.

ms609:
Please direct any further discussion on this topic to the related [issue](https://github.com/ms609/citation-bot/issues/177).

PR #171, opened 2017-09-21
Make inline slow--hope it works

GlazerMann:


GlazerMann:
Does this make sense?

gadgetapi.php
```diff
@@ -14,6 +14,7 @@
 //Expand text from postvars
 $page = new Page();
 $page->text = $originalText;
+$_REQUEST["slow"] = 1;
```
ms609: You can't set $_REQUEST variables, unfortunately.  
To do this, visit citationbot.com/rest-of-url.html?**slow=1**

GlazerMann:
I have now put in a request on media wiki to pass slow

ms609:
Oh, I think I understand what you are trying to do now.  The alternative would be to set `$slow_mode = 1` (is that what the variable is called?) in `gadgetapi.php`.

GlazerMann:
Not that easy actually.  I have asked Wikipedia to edit the actual tool that calls the bot so that slow comes from the client.   I have copied the source into my common.js and verified that it works that way

GlazerMann:
Fixed on wikipedia

https://en.wikipedia.org/w/index.php?title=MediaWiki%3AGadget-citations.js&action=historysubmit&type=revision&diff=802331109&oldid=798719114

PR #172, opened 2017-09-25
catch rookie mistakes with usage of website

GlazerMann:


Template.php
```diff
@@ -444,6 +444,19 @@ public function add_if_new($param_name, $value) {
   }
 
   protected function get_identifiers_from_url() {
+    if ($this->blank('url') && !$this->blank('website')) {  // No URL, but a website
+      $url = trim($this->get('website'));
+      if (strtolower(substr( $url, 0, 4 )) !== "http" ) {
+        $url = "https://" . $url; // Try it with http
```
ms609: Could we go a step further with a regexp to catch typos such as `website = ttp:/web.com`?  (Might not be necessary in enough cases to justify)
Also is it worth testing `http` as well as `https` to catch sites that lack SSL?

Template.php
```diff
@@ -444,6 +444,19 @@ public function add_if_new($param_name, $value) {
...
+      }
+      if (filter_var($url, FILTER_VALIDATE_URL,FILTER_FLAG_HOST_REQUIRED |  FILTER_FLAG_PATH_REQUIRED ) ) { // is a valid http(s) URL with a full hostname and path of some type
+        $this->set('url',$url);
+        $this->forget('website');
```
ms609: Rather than using "set" and "forget", it would be better to rename the existing parameter; this will preserve its position in the citation and its formatting/spacing.

Template.php
GlazerMann: Improving

Template.php
GlazerMann: will do

GlazerMann:
I think this now much better.   That regex is insane, but necessary if we are going to claim that a website is actually a URL.

PR #174, opened 2017-09-29
Better handling of year/date alias

GlazerMann:


GlazerMann:
The SICI data path sets year= even if date=is already set.   This was from a bug report.

The removing of "in press" when adding a date, is my own discovery.

PR #175, opened 2017-09-29
remove in press when setting date

GlazerMann:


ms609:
Could we also add aliases for 'in press', such as 'forthcoming', 'in the press'?

GlazerMann:
I have no idea why this is failing to build

ms609:
Looking at the Travis build report, it looks like the problem is a call to an undefined `unset` method in the `Template` class.  Perhaps you mean `forget()`?

PR #176, opened 2017-09-29
handle aliases when using SICI data

GlazerMann:


Template.php
```diff
@@ -981,13 +981,13 @@ public function expand_by_pubmed($force = FALSE) {
 
   protected function use_sici() {
     if (preg_match(siciRegExp, urldecode($this->parsed_text()), $sici)) {
-      if ($this->blank($journal, "issn")) $this->set("issn", $sici[1]);
+      if ($this->blank($journal,"journal")) $this->add_if_new("issn", $sici[1]);
```
ms609: `$this->blank($journal,"journal")` looks like a hangover from a previous code formulation.
`$journal` is not set in the use_sici scope.  Perhaps it formerly contained the string `journal` or `periodical` depending on which parameter was set in the template?
`$this->blank` takes one parameter, not two.
Probably the solution is just to delete the first `$journal`?

PR #182, opened 2017-09-29
more tests

GlazerMann:


tests/phpunit/TemplateTest.php
```diff
@@ -182,6 +182,10 @@ public function testMisspeltParameters() {
     $this->assertNotNull($expanded->get('journal'));
     $this->assertNotNull($expanded->get('pages'));
     $this->assertNotNull($expanded->get('year'));
+      
+    $text = "{{Cite journal|pp. 1–5}}";
```
ms609: Is this redundant to `testUseUnusedData` (line 36)?

tests/phpunit/TemplateTest.php
```diff
@@ -204,4 +208,30 @@ public function testErrantAuthor() {
...
+    $this->assertNull($expanded->get('date'));
+    $this->assertEquals('1999',$expanded->get('year'));
+
+    $text = '{{cite journal|pmid=9858585|year=in press}}';
+    $expanded = $this->process_citation($text);
+    $this->assertNull($expanded->get('date'));
+    $this->assertEquals('1999',$expanded->get('year'));
+  }
+  
+  public function testSICI() {
```
ms609: Does `testSICI` test anything that is not covered in `testSiciExtraction()` (line 146)?

tests/phpunit/TemplateTest.php
```diff
@@ -182,6 +182,10 @@ public function testMisspeltParameters() {
     $this->assertNotNull($expanded->get('journal'));
     $this->assertNotNull($expanded->get('pages'));
     $this->assertNotNull($expanded->get('year'));
+      
+    $text = "{{Cite journal|pp. 1–5}}";
+    $expanded = $this->process_citation($text);
+    $this->assertNotNull($expanded->get('pages'));
```
ms609: Is it worth using `AssertEquals('1–5', $expanded->get('pages');`?

GlazerMann: Will do

tests/phpunit/TemplateTest.php
GlazerMann: Missed that one

tests/phpunit/TemplateTest.php
GlazerMann: This is for that bug where the first element gets deleted.  Don't want it coming back

PR #185, opened 2017-09-30
More test cases.  RIS, EndNote, etc

GlazerMann:


GlazerMann:
This does not work yet.  Probably bugs in Bot

GlazerMann:
Do NOT commit this just yet.  Some other fixes are submitted.  Also, I need to check output.

GlazerMann:
Do not commit until error free

PR #188, opened 2017-10-01
Test -  do not commit

GlazerMann:
This forces slow mode testing


GlazerMann:
Just do a null edit on this to find errors in slow mode.  
Some tests fail because slow mode finds more things.

ms609:
I've changed how SLOW_MODE it set (now a global variable) so we can use it whilst testing.  I've turned it on by default in templateTest.php.

PR #191, opened 2017-10-01
sometimes neither page nor pages is set

GlazerMann:


ms609:
If neither is set, then `$this->get('page')` will return NULL; is there a case where your re-writing of the code will have a different outcome?

ms609:
PS if you're making a pull request for each change to see whether it throws an error in Travis, you might be finding the debugging quite a frustrating process!  
What I do is use Github Desktop to download the source files to my machine; then I can run the tests locally (by installing php, adding the directory containing `phpunit` to the `PATH` system environment variable in Windows, then using the command prompt (`cmd`) to navigate to the folder containing the code files and running the command `phpunit --bootstrap ./expandFns.php tests/phpunit/templateTest.php`). Hopefully you'll find that a little less painstaking...

GlazerMann:
Pages() is good idea

PR #201, opened 2017-10-02
Add support for outdated JSTOR URLS

GlazerMann:


ms609:
Great call!  Are you able to add a corresponding test case to the pull request?

GlazerMann:
Done as a separate pull.   That’s the best I can do from my phine

GlazerMann:
I should note that I did verify that jstor does not rewrite bad urls, so we don’t have to worry about that

PR #205, opened 2017-10-02
Convert ISBN10 to ISBN13 and change add to add_if_new

GlazerMann:


ms609:
Unless there are circumstances when one might want to add an ISBN without undergoing this check, my suggestion would be to call the function once from within add_if_new, rather than each time add_if_new is called.  One fewer thing to remember when adding calls in the future, it catches cases where the parameter name is now hard-coded (e.g. conversion of ID parameter), and it keeps all the handy format checks in one place.

GlazerMann:
I will do that

GlazerMann:
This fails the quality control because add_if_new correctly sets the "year" parameter instead of the "date" parameter.

GlazerMann:
This is now called in add_if_new
Also several adds are now add if new
Finally called in tidy() which should fix existing ones-you can decide if that is wise

GlazerMann:
I think this is done now.

Template.php
```diff
@@ -2343,4 +2352,22 @@ public function modifications ($type='all') {
   public function is_modified () {
     return (bool) count($this->modifications('modifications'));
   }
+  
+  public function isbn10Toisbn13 ($isbn10) {
+       $isbn10 = trim($isbn10);  // Remove leading and trailing spaces
+       $isbn10 = str_replace(array('—','―','–','−','‒'),'-', $isbn10); // standardize dahses : en dash, horizontal bar, em dash, minus sign, figure dash, to hyphen.
+       if (preg_match("/[^0123456789Xx-]/",$isbn10) === 1)  return $isbn10;  // Contains invalid characters
```
ms609: Need to escape the hyphen.
`"/[^0123456789Xx\-]/"`
Or more concisely
`"/[^0-9Xx\-]/"`

For consistency I'd also suggest using `~` instead of `/` as the regular expression delimiter.

ms609: Are you able to add a test case too, please?

PR #208, opened 2017-10-02
Journal Italics - add test

GlazerMann:


GlazerMann:
This is a bug in Dev branch

ms609:
Looks like it thinks it's a species, from the taxon matching regexp in format_title_text [sp?].  Perhaps we need to drop those lines in format_title_test, only call them if it's a new title, or add a capitalization exclusion for that particular journal.

GlazerMann:
It seems to be any internal capitalization causes this.

GlazerMann:
$p->val = format_title_text(title_capitalization($p->val, FALSE, FALSE));
    

In tidy()

PR #211, opened 2017-10-03
Journal Italics - use option

GlazerMann:


GlazerMann:
This depends upon the other pull that changes the function

PR #214, opened 2017-10-04
year and date are equally valid sometimes

GlazerMann:


GlazerMann:
This code works too, and avoids if/else/if_null stuff:
$this->assertEquals('2002', $expanded->get('year') . $expanded->get('date') );
thoughts?

ms609:
I like the concision, but it could cause a problem if both date and year are set (to different values).
My suggestion would be for a new public function from within the Template class that uses your original if (year) year else date syntax.

PR #219, opened 2017-10-06
ISBN test cases

GlazerMann:


GlazerMann:
None of these test NEW citations.  They all count on a call to tidy().  Google seems to send ISBN13 always these days.

PR #226, opened 2017-10-10
Do not flag placeholders as bad DOIs

GlazerMann:
Do not flag placeholders as bad DOIs

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -315,7 +315,7 @@ class Template extends Item {
 
         if ($this->blank("last$auNo") && $this->blank("author$auNo")
           && $this->blank("coauthor") && $this->blank("coauthors")
-          && strpos($this->get('author') . $this->get('authors'), ' and ') === FALSE
+          && strpos($this->get('author') . $this->get('authors'), ' and ')   FALSE
           && strpos($this->get('author') . $this->get('authors'), '; ') === FALSE
           && strpos($this->get('author') . $this->get('authors'), ' et al') === FALSE
         ) {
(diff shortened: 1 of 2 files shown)


ms609:
To reduce the risk of false positives, would it be better to look for a match with the placeholders themselves?
`sprintf(PLACEHOLDER, '\d+')` will convert a placeholder to a regular expression.  Or it shouldn't be too tricky to write a regexp that would catch the exisitng placeholder formats.

ms609:
Looking at this again, it looks like what we really want to do is to strip out any placeholders at the point of reading the DOI.  So a DOI reading `doi = 10.1038/nature12345<!--here is a comment-->` should have the comment removed before being treated normally.

GlazerMann:
0. trim()
1.  Look for # # ........... # # start and end.
2.  Remove and save it to a temporary string AND remember if start or end
3.  Look for another one, if another one then abort.  This is paranoia.
4.  Process DOI
5.  Re-add # # ............ to front and back as needed.

ms609:
I was thinking more along the lines of
```php
$doi = preg_replace_all($placeholder_format, '', $this->get('doi'));
if (doi_is_broken($doi)) {
  $this->set('doi_broken_date');
}
```

GlazerMann:
The Bot documentation states that if the doi parameter is comment (and just a comment), then bot will not fill it.  That way you can block an invalid DOI.  So, should we remove that documentation?

GlazerMann:
All the changes are in one location now.  A single function that was already being used get_without_comments.

GlazerMann:
Interesting.  The template placeholder does not get Un-Escaped

GlazerMann:
This is ready to go

GlazerMann:
We are good to go!!!!!!

GlazerMann:
Thank you for explanation of how to properly test this as a page, and not a template

GlazerMann:
Please push this on in now.

PR #229, opened 2017-10-13
uncomment mysql.login inclusion

kaldari:
This is necessary for doibot.php to run properly. It's been manually patched on citations-dev.


ms609:
I'm merging this into the `release` branch, rather than `master`, as uncommenting the line will cause the continuous integration tests to fail.  The `release` branch contains a couple of other relevant modifications (e.g. not including 'dummy' login details) and is intended to be the branch that is 'hosted' on WMFlabs servers.  I hope that this does the job, but do let me know if not!

PR #237, opened 2017-10-13
Open Access might be PMC, etc.

GlazerMann:
Fixed open access code adding URL that will be updated by later bot runs or duplicates PMC etc.


GlazerMann:
Opinions on usefulness of this?

PR #238, opened 2017-10-13
update PMC URL - tests are failing

GlazerMann:
Fixes PubMed changing URL


GlazerMann:
This test only failed because http://adsabs.harvard.edu/cgi-bin/abs_connect?data_type=XML&doi=10.1103%2FPhysRevD.78.081701 got angry with us polling too many times

ms609:
I've rejigged this test in a different way to reflect other changes - it'll now add a PMC instead of a URL.

PR #241, opened 2017-10-14
Formatting of titles and journal names

GlazerMann:
Fixes strange re-formatting of titles with wiki text
Fixed attempts to find italics in journal titles


ms609:
Your test cases have uncovered a deeper problem, which is that we're not distinguishing new text drawn from external databases from manually-entered text already within citations.  I've renamed format_title_text to make this distinction clearer, and changed when it's called; hopefully this provides a simple way of addressing the problem.

PR #244, opened 2017-10-14
parse_text() is called many places

GlazerMann:
parse_text() restarts a zero, and that seems like a bad idea since it is called all over the place

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -107,7 +107,7 @@ class Template extends Item {
     process_real();
     $this->text = $this->replace_templates($this->text);
   }
-  private function process_real() {
+  public function process_real() {
     switch ($this->wikiname()) {
       case 'cite web':
         $this->use_unnamed_params();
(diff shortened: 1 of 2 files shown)


GlazerMann:
Obviously not getting unwrapped either enough 😲

Might also want two dimensional array where encoding and decoding is a stack.

GlazerMann:
I have a test case, but no solution.

ms609:
Wouldn't the internal comments have been replaced by placeholders when the page was parsed?  `process_page` should be used instead of `process_citation` here.  Unfortunately that makes the test cases a little more complicated.
```php
$text = '{{Cite journal|url={{This is not real}} | doi = {{I am wrong}} |jstor= {{yet another bogus one }}  }}';
       $expanded = $this->process_page($text);
       print $expanded->parsed_text();
```

GlazerMann:
Comments seem to be okay.   It is templates that do not make it out alive.

Happens with real pages too.

ms609:
Seems to be passing okay in a hasty test but I'll look into this properly anon.

GlazerMann:
Closing and pointing to other open pull

PR #248, opened 2017-10-18
Adding missing space and period

kaldari:


GlazerMann:
Should apply to all branches. 👍

ms609:
Could you create a separate patch for the `master` branch so the change doesn't get overwritten when I next update the release branch? Ta.

PR #251, opened 2017-10-20
Fix Gadget API-Do not echo stuff to wikipedia

GlazerMann:
Fixes Gadget API.

Bot sends this when processes page with simply "MyText":
[15:36:43] Processing page ...
{"expandedtext":"MyText","editsummary":"\/* Testing *\/  | [[WP:UCB|Assisted by Citation bot]]"}
Leads to error on wiki side because of processing page line is not valid json:
SyntaxError: JSON.parse: expected ',' or ']' after array element at line 2 column 4 of the JSON data

I was confused about cause.  Turns out that HTML had nothing to do with it.

This turns on the eating of the buffer sooner.

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -564,7 +564,7 @@ ER -  }}';
            'editsummary' => $editSummary,
        );
        $expanded = json_encode($result);
-       $this->assertEquals('{"expandedtext":"This is a page. {{Cite web|url=http:\/\/apple.com\/phones\/buy_android}}. Indeed it is","editsummary":"I made this page | [[WP:UCB|Assisted by Citation bot]]"}',$expanded);
+       $this->assertEquals(base64_encode('{"expandedtext":"This is a page. {{Cite web|url=http:\/\/apple.com\/phones\/buy_android}}. Indeed it is","editsummary":"I made this page | [[WP:UCB|Assisted by Citation bot]]"}'),base64_encode($expanded));
    }
   /* TODO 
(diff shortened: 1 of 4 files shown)


GlazerMann:
The existing code does not work because defines do not get passed in.

GlazerMann:
I think this is now done.

ms609:
This brings to my attention the lack of any test cases for `gadgetapi.php` ... at a minimum, is it possible to create a case (probably in a new file gadgetapiTest.php in the testcases folder) that checks that the output of the page is valid JSON in a couple of instances?

GlazerMann:
>  is it possible to create a case that checks that the output of the page is valid JSON in a couple of instances.

I do not think that this is straight forward.  We would need to:
1.  setup PHP's webserver to run the gadgetapi.php page.  See:   http://symfony.com/doc/current/setup/built_in_web_server.html
2. query this webserver with an http request and save the output
3. parse this data as json and verify it

GlazerMann:
Getting the white space right was a pain in the JSON.  We now have a "test" of sorts.

GlazerMann:

tests/phpunit/TemplateTest.php
```diff
@@ -564,6 +564,30 @@ public function testConvertJournalToBook() {
        $this->assertEquals('cite book', $expanded->wikiname());
    }
     
+   public function testGadgetAPI() {
```
ms609: What we need to test, if anything, is that nothing is echoed when gadgetapi.php is run (except the final output).  As you say, the functionality itself is tested in TemplateTest.php and elsewhere.  It's not clear that this tests that.

PR #256, opened 2017-10-23
echo is slightly faster than print

GlazerMann:
Fixes mixed us of print and echo.
Print has a return code, which we do not use most of the time.
The return code is used in "do this or print" type code.


ms609:
Not merging this straight away as my convention has been to use `echo` for things that are meant to be there, and `print` for temporary commands that are intended for debugging.  I want to update the docs with this note, and check carefully which category each of these prints falls into - it's possible that some should be replaced with `warn` rather than `print`.

PR #257, opened 2017-10-24
echo vs. print.  die vs. exit

GlazerMann:
Never use die or print in production code
All die and print removed


GlazerMann:
I think I got all of them.  Some are a matter of opinion

GlazerMann:

ms609:
In practice I think the distinction is too arbitrary - I think perhaps we should use `echo` and `exit` throughout.  I think this is actually closer to my original intention in any case -- I use `print` and `die` to throw debugging messages, so can search for these terms before making a commit to make sure I've not left any in.  Hope that's okay?

GlazerMann:
Are any of these categorized wrong?

GlazerMann:
I believe that this is now every die and print statement.

PR #258, opened 2017-10-24
Define HTML_OUTPUT as Boolean 

GlazerMann:
HTML_OUTPUT is defined all over the place.  It is only used in one place, so define it here.

This assumes that the usage of "-1" in the Gadget API is no longer done


GlazerMann:

ms609:
Won't this change `HTML_OUTPUT` from `FALSE` (-1) to `TRUE` for all entry points except `doibot.php`?  The idea is that HTML_OUTPUT is set in calling scripts before ExpandFns.php is included for the first time.

GlazerMann:
Is HTML_OUTPUT defined anywhere.  It is not defined in diobot.php anymore, since the grabbing of the output is now done there, instead of flagged with -1.

PR #259, opened 2017-10-25
Do not reformat Wiki-linked journal names

GlazerMann:
Fixes existing wiki-linked names getting formatted.  We just leave as-is.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -428,7 +428,11 @@ class Template extends Item {
       
       case 'volume':
         if ($this->blank($param_name)) {
-          if (in_array(strtolower($this->get('journal')), HAS_NO_VOLUME) === TRUE ) {
+          $temp_string = strtolower($this->get('journal')) ;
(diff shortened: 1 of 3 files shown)


GlazerMann:
@ms609 When you have a chance.  Please look over the outstanding pulls and commit or comment.  I hope that they are all "outstanding" in both senses of the word.

ms609:
My stance on this one is that there's never a time that a journal name should be wikilinked! This may be reflected in the template documentation on Wikipedia, or you could push for it... this must be an edge case!
Does the code handle piped links such as `[[ZooKeys (journal)| ZooKeys]]`?

GlazerMann:
It does not handle that case.
I agree about wiki links to journals
Now that you mention it.  We should simply not touch them since who knows what’s in them.
I will update patch

GlazerMann:
This now just gives up and does not try to format them at all if wiki-linked.  

Why do people do this?  I can think of one and only one reason to do that.  The journal is old and discontinued and thus has no ISSN, no DOI, no website, etc..  Then a wikilink to the journal is a good idea; and only then!

PR #261, opened 2017-10-29
Forget dx.doi.org URLS when getting DOIs from them

GlazerMann:
Fixes inconsistency in using URLS.  
When we find a DOI, get rid of URL and add DOI.


ms609:
Does this handle cases where the URL provides access to a freely accessible copy of the text, whereas the DOI link goes to a paywall?  In these cases the URL should be retained.

PR #265, opened 2017-10-31
Rebasing soon

GlazerMann:
One API key for testing servers.  One for productions.  They are both restricted.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1243,11 +1243,6 @@ class Template extends Item {
   protected function expand_by_google_books() {
     $url = $this->get('url');
     $isbn= $this->get('isbn');
-    if($isbn) { // Can fake it, if no URL or non-google URL
-      if (!$url || ! preg_match("~books\.google\.[\w\.]+/.*\bid=([\w\d\-]+)~", $url, $gid) ) {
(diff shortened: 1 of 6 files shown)


GlazerMann:
Overly optimistic today.  Almost done/

ms609:
Before rolling this out we'll probably need an API key
https://developers.google.com/books/docs/v1/using#auth

My current job (not that I've had any time to progress it for a while!) is to use OAuth to log the bot in to the WP servers, which might (?) help with authorizing requests once it's working.

GlazerMann:
Need Google API key.  True.
URL needs this added
&key=yourAPIKey

GlazerMann:
Obviously, the empty string “key” needs to be replaced with a real one

GlazerMann:
The Google API key is set to only work on tools.wmflabs.org has address 208.80.155.131

ms609:
I wonder how we can set up a test suite for this; we'd need to configure the API to work on Travis.
With other APIs I have one key that I put in the credentials folder on GitHub, and a separate key that I use for production (which I upload directly to the server and keep private).  Is this a possible solution?

GlazerMann:
To get the API to work on travis, I just need to make another key that only works on travis IP addresses

GlazerMann:
I think this ready to merge.

PR #266, opened 2017-11-01
add jstor support using Citoid

GlazerMann:
Fixes missing jstor support
Includes test case

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -103,7 +103,6 @@ class Template extends Item {
   }
 
   public function process() {
-    ini_set('user_agent','The Wikipedia Citation Bot');  // This gets passed in HTTP headers
     switch ($this->wikiname()) {
       case 'cite web':
         $this->use_unnamed_params();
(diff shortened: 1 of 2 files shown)


GlazerMann:
Not fully baked yet.  But, open for comments

GlazerMann:
incompleteBook() is not actually used at this time.

GlazerMann:
Jstor has added more stuff.  The SICI link now points to 25 articles.  I will fix.   That is why SICI links need fixed

GlazerMann:
https://github.com/zhaofengli/refill/blob/11ef3b3b35bcd686ffd5aacb9c2dafe9dbf8e5e1/src/Reflinks/LinkHandlers/CitoidLinkHandler.php

This points to reFill parsing code.  Just found it.

ms609:
Great job, lovely to have JSTOR supported again!
Could you add a test case that adds authors to a journal too, please? These parameters are often particularly irksome so worth checking that they are covered.

PR #273, opened 2017-11-05
incomplete() is different for books

GlazerMann:

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -200,7 +200,17 @@ class Template extends Item {
     }
   }
 
-  protected function incompleteJournal() {
+  protected function incomplete() {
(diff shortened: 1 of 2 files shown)


ms609:
Good to make this distinction, but I must admit that my first thought from the revised name was that it checked whether the journal parameter (or equivalent) was incomplete.  
That said I'm struggling to quickly think of a clearer alternative name...
My feeling is that we use underscores in naming functions, rather than camelCase (except in the largely obsolete DoiTools.php) -- would you mind checking, and perhaps updating our style guide accordingly?
Thanks!

GlazerMann:
I HATE camel case.  I can no believe I used it

PR #275, opened 2017-11-05
Revert "de-encode SICI"

ms609:
Come to think of it, it could be worth testing the encoded SICI as well as the decoded one, to be confident that the decoder works correctly...


GlazerMann:
Add this one:
https://www.jstor.org/sici?sici=0080-4649(1969)173:1031<235:TPOBAG>2.0
going to 75817
That way we check ecoded and non-endcoded

PR #290, opened 2017-11-10
Store publisher endings in an array, not a bunch of if statements

GlazerMann:
Fixes the ever growing list of publisher endings getting out of control in code base by making it an array of publisher endings.  Because we use "explode()" we know that there is a space in front of them, so we do not have to worry about some ones last name being "Splinc" being caught as "inc".  If your last name happens to be "magazines" or "inc", then sorry.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1389,7 +1389,7 @@ class Template extends Item {
       foreach ($xml->dc___creator as $author) {
         if( in_array(strtolower($author), BAD_AUTHORS) === FALSE) {
           $author_parts  = explode(" ",$author);
-          $author_ending = ' ' . end($author_parts);
+          $author_ending = end($author_parts);
           if( in_array(strtolower($author),       AUTHORS_ARE_PUBLISHERS        ) === TRUE ||
               in_array(strtolower($author_ending),AUTHORS_ARE_PUBLISHERS_ENDINGS) === TRUE) {
             $this->add_if_new("publisher" , (str_replace("___", ":", $author)));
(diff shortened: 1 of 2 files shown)


GlazerMann:
Failed because of : 
1) TemplateTest::testCitoidJstorExpansion

Failed asserting that two strings are equal.

--- Expected

+++ Actual

@@ @@

-'Are Helionitronium Trications Stable?'

+'Are Helionitronium Trications Stable? On JSTOR'

Added a pull for this.

Template.php
```diff
@@ -1388,15 +1388,10 @@ protected function google_book_details ($gid) {
...
-             substr(strtolower($author),-4)  === " inc" || 
-             substr(strtolower($author),-5)  === " inc." ||
-             substr(strtolower($author),-10) === " magazines" || 
-             substr(strtolower($author),-6)  === " press" ||
-             substr(strtolower($author),-6)  === " books" ||
-             substr(strtolower($author),-11) === " publishing" ||
-             substr(strtolower($author),-11) === " publishers" ||
-             substr(strtolower($author),-12) === " corporation") {
+          $author_parts  = explode(" ",$author);
+          $author_ending = ' ' . end($author_parts);
```
ms609: could we do without this line by removing spaces from strings below?

constants.php
```diff
@@ -29,6 +29,7 @@
 // Use lower case for all of these, and then compare to a lower cased version
 const BAD_AUTHORS = array("unknown","missing");
 const AUTHORS_ARE_PUBLISHERS = array(); // Things from google like "hearst magazines", "time inc", "nielsen business media, inc" that the catch alls do not detect
+const AUTHORS_ARE_PUBLISHERS_ENDINGS = array(" inc."," inc"," magazines"," press"," publishing"," publishers"," publishing"," books"," corporation");
```
ms609: spaces would have to be deleted here.

Template.php
```diff
@@ -1388,15 +1388,10 @@ protected function google_book_details ($gid) {
...
+          if( in_array(strtolower($author),       AUTHORS_ARE_PUBLISHERS        ) === TRUE ||
+              in_array(strtolower($author_ending),AUTHORS_ARE_PUBLISHERS_ENDINGS) === TRUE) {
```
ms609: Another point for the style guide: commas should always be followed by a space, if that's okay.

Template.php
GlazerMann: Done!

Template.php
GlazerMann: Done!

constants.php
GlazerMann: Done!

PR #291, opened 2017-11-10
citoid sometimes adds "on jstor" to titles

GlazerMann:
Fixes random "on jstor" added to titles.


GlazerMann:
And it looks like sometimes it does not grab volume data.  :-(

ms609:
Fixes #292  -- you beat me to it!

PR #296, opened 2017-11-10
Wrong function called — empty vs blank

GlazerMann:
Fixes empty vs blank


GlazerMann:
I really need to remember empty vs blank

PR #297, opened 2017-11-10
Deal with Citoid failures

GlazerMann:
Based on the XML data: sometimes citoid just reads it as a website, and misses that it is a journal
Code now has jstor URL that does not redirect (adds on &seq=1 kind of stuff).  I think this should greatly reduce the failures.
Code now checks for "not found" as a title.
If failure, code retries with slightly different URL to make sure Citoid does not cache answers
Test now complains -- if you read logfile -- but does not bomb out.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1138,14 +1138,16 @@ class Template extends Item {
     if ( isset($data[0]->{'title'})) {
       $the_title_data = trim($data[0]->{'title'});
       if (strtolower(substr($the_title_data,-9)) === ' on jstor') {
-         $the_title_data = substr($the_title_data, 0, -9); // Citoid did not pick up that it was a journal.  Nothing else is probably found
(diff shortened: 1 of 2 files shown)


ms609:
Have you any idea why citoid is failing?  Is there any way that we can respond to failure - by re-trying the request, for example?

GlazerMann:
I have looked at XML data.   Sometimes it just reads it as a website, and misses that it is a journal

GlazerMann:
Retry code added.   It’s tipped off by jstor in title.

GlazerMann:
Code now has jstor URL that does not redirect (adds on &seq=1 kind of stuff)
Code now checks for "not found" as a name.

GlazerMann:
Oh yeah.  And the retry URL is slightly different to make sure Citoid does not cache answers

GlazerMann:
I think that it should work much better now

ms609:
Looks great!

PR #298, opened 2017-11-11
Better operator choice

GlazerMann:
It does not matter in this case, but good coding means we should swap it.
$this = true;
$that = false;
$truthiness = $this and $that;
$truthiness is true in the above case. :-( because "=" evaluates before "and"
And now I present you will the pickiest pull request ever........


GlazerMann:
I did search the dev branch and this is the only "and"/"or" we use.

PR #299, opened 2017-11-11
Eliminate all class extensions for performance and simplicity

GlazerMann:
This abstract class used to do a lot, now it does almost nothing.
Code is simpler this way.  We never cast from one to another so common base was not used.
Also declare all our classes final, since that makes php a nano second faster.


GlazerMann:
That awkward moment when you realize that you submitted each file change seperately and overwhelmed your Travis allocation 🤭

PR #303, opened 2017-11-12
verify get_prefix_index does not return wrong stuff

GlazerMann:
Pass back empty array instead of NULL, when code fails - bug fix.
Also add test to verify that get_prefix_index does does not pass back anything if you ask for something that there are zero of.  Existing test just makes sure you get the item you want; while this test makes sure you do not getting extra stuff back.
This is from just looking for abc[]= stuff that might end up empty


GlazerMann:
This actually died because arxiv returned nothing for one Call.  The build actually works

PR #304, opened 2017-11-12
Template expansion does not really work

GlazerMann:
Processing as a page leads to an illegal array access. 
If you set php to ignore errors, then the final output does not actually expand all the templates.
In case you are wondering how found: I set expand template to also expand as a page and this was the only one with a access error.  I verified no output, just errors.
This explains why this feature works in tests suite but not in the real world.

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -115,7 +115,9 @@ class Page {
     }
 
     for ($i = 0; $i < count($templates); $i++) {
+      $templates[$i]->extract_my_templates();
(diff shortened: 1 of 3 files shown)


GlazerMann:
the source code changes have all failed to work

GlazerMann:
That awkward moment when you submit too many change requests and zap Travis allocation

GlazerMann:
I am not sure how to fix this.   But we have the testing now

PR #305, opened 2017-11-13
Template expansion does not really work

GlazerMann:
Processing as a page() leads to an illegal array access.
If you set php to ignore errors, then the final output does not actually expand any templates.
This explains why this feature works in tests suite but not in the real world.
The template needs add other templates passed to it (AS A POINTER)
Also use index instead of number

This is because we parse the whole page and when we jump into a template with process(), there are no templates left internally.

I have chopped out a couple parts of this and submitted them as separate pulls, if you want to think about this.

The only thing I am not 100% about is my choice in the  __toString() function.

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -116,8 +116,7 @@ final class Page {
 
     for ($i = 0; $i < count($templates); $i++) {
       if (count($templates[$i]->internal_templates) !==0) {
-        echo "Attemping to set internal_templates twice";
-        return FALSE;
+        exit("Attemping to set internal_templates twice");
       }
       $templates[$i]->internal_templates = &$templates;   
       $templates[$i]->process();
--- a/Page.php
+++ b/Page.php
@@ -115,9 +115,7 @@ final class Page {
     }
 
     for ($i = 0; $i < count($templates); $i++) {
-      if (count($templates[$i]->internal_templates) !==0) {
-        exit("Attemping to set internal_templates twice");
-      }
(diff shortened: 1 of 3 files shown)


GlazerMann:
This fails also as a template:
{{cite book | id={{arxiv|astr.ph|1234.5678}} {{arxiv|astr.ph|1234.5678}} }}
Since we treat identical templates that same and then assume that indexes to placeholders and template number will match.

GlazerMann:
Two issues here.   One fixed

GlazerMann:
Holy cow it work!!!!

GlazerMann:
I removed these tests from the PULL:

+  public function testId2ParamPage() {
+      $text = '{{cite book |id=ISBN 978-1234-9583-068, {{ol|12345}} }}';
+      $expanded = $this->process_page($text);//caused undefined access
+  }
+    
+  public function testId2ParamPageSuppressErrors() {
+      $text = '{{cite book |id=ISBN 978-1234-9583-068, {{ol|12345}} }}';
+      error_reporting(E_ALL^E_NOTICE);
+      $expanded_page = $this->process_page($text);
+      error_reporting(E_ALL);
+      $expanded_cite = $this->process_citation($text);
+      $this->assertEquals($expanded_cite->parsed_text(), $expanded_page->parsed_text());
+  }

Because they worked, and the page() vs. template() distinction when running tests goes away in another pull

PR #313, opened 2017-11-14
?> is supposidly a bad idea.

GlazerMann:
pure php files are not supposed to end 🙄


GlazerMann:
I have requested a bigger Google Books allocation

PR #316, opened 2017-11-15
Fix all tests

GlazerMann:
Fixes tests not matching real world
You have to run tests as a Page() for them to match what happens on Wikipedia.
After running them as page use Template() to access them
Some tests fail because of this, but that is good to know.

This is included in the big pull but it can stand alone

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -24,8 +24,8 @@ final class TemplateTest extends PHPUnit\Framework\TestCase {
   
   protected function process_citation($text) {
     $template = new Template();
-    $template->parse_text($text);
-    $template->process();
+    $page = $this->process_page($text);
+    $template->parse_text($page->parsed_text());
     return $template;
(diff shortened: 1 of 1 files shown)


tests/phpunit/TemplateTest.php
```diff
@@ -24,8 +24,8 @@ protected function tearDown() {
   
   protected function process_citation($text) {
     $template = new Template();
```
ms609: Do we need a `$page= new Page()`?

GlazerMann: that’s an excellent question.   The new Page() is made in process_page()
Here is the code with subroutine kind of inlined 

    $template = new Template();
    $page = $this->process_page($text);
         $page = new Page();
         $page->parse_text($text);
         $page->expand_text();
    $template->parse_text($page->parsed_text());
    return $template;

GlazerMann:
I have made expand_citation stand alone.

PR #333, opened 2017-11-18
add tests of Page() functions

GlazerMann:
Add code coverage


GlazerMann:
This build actually worked, we just slammed bibcode people with too many requests at once

PR #337, opened 2017-11-20
Should be new Page() for each new page

GlazerMann:
Page() initializes things in its construction phase


GlazerMann:
$text, $title, $modifications are all reset.  I was wrong

PR #351, opened 2017-11-25
Add more test cases

GlazerMann:
Fixes lack of code coverage


tests/phpunit/wikiFunctionsTest.php
```diff
@@ -188,5 +188,33 @@ public function testCurlSetup() {
     curl_setup($ch, $url);
     $this->assertNull(NULL); // Just looking for code coverage and access of unset variables, etc.
   }
+
+  public function testPageIsRedirect() {
+    $page = new page();
+    $page->title = 'WP:UCB';
+    $this->assertEquals(1, $page->is_redirect()[0]); // Different code than non-class function is_redirect($text)
+  }
+  public function testPageTextFromTitle() { // Not a great test. Mostly just verifies no crashes in code
```
ms609: I put in a similar test in my codecov branch, which points to a page I'd set up on Wikipedia to have a known (static and edit-protected) page content, under _User:Bot testing account_ or something similar.  This allows testing that the page contents are as expected.  I'll see if I can dig it out when I get the chance.

tests/phpunit/wikiFunctionsTest.php
```diff
@@ -188,5 +188,33 @@ public function testCurlSetup() {
...
+  public function testEditSummary() {  // Not a great test. Mostly just verifies no crashes in code
+    $page = new Page();
+    $text = "{{Cite journal|pmid=9858586}}";
+    $page->parse_text($text);
+    $page->expand_text();
+    $this->assertNotNull($page->edit_summary());
+  }
+  
+  public function testMathInTitle() {
+    // This MML code comes from a real CroffRef search of DOI 10.1016/j.newast.2009.05.001
```
ms609: Owch! (-:

GlazerMann: No kidding.  That is why I explcitly include the MML here.  I do not want CrossRef changing this on us.  We really need this test.

tests/phpunit/wikiFunctionsTest.php
GlazerMann: I am an Idiot.  We already test this function.  👎   I have removed it from this pull.  I say idiot, since I am the person who submitted the previous pull.

PR #357, opened 2017-11-26
fix templates within templates

GlazerMann:
Within Page() all templates are replaced. Thus there are no internal templates. This code sets the internal template list to match the global list of templates.

IT ACTUALLY WORKS 

I had to create a whole slew of tests to make sure this did not break stuff.

When merging this you definitely want to chose the forget all intermediate modifications option


ms609:
This has blown my mind.  Amazing job!

PR #366, opened 2017-11-30
Some bibcode data is bad

GlazerMann:
Pages of "arXiv:1711.02260" is returned along with volume of "1711" and journal of  "arxiv e-prints".

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1027,7 +1027,7 @@ final class Template {
             $this->add_if_new('journal', $journal_string[0]);
           }          
         }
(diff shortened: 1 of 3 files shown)


Template.php
```diff
@@ -418,6 +418,7 @@ public function add_if_new($param_name, $value) {
         return FALSE;
         
       case "periodical": case "journal":
+        if (strtolower($this->get('journal')) === 'arxiv e-prints') $this->forget('journal'); // Update to real data
```
ms609: Should we instead check that the journal title is not any of BAD_TITLES?

GlazerMann: Excellent point.  Will do.   I will rue the day when a hippie starts the open access data journal “Unknown”

PR #373, opened 2017-12-02
Fixes parallel builds crashing in APIs

GlazerMann:
Fixes parallel builds (I hope).
Add Zzzzzzzzzz
This should also speed up 7 builds and overloaded servers.  Shorten the code, test, fix loop.
Also put sleep(1) between google API calls.

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -575,6 +575,7 @@ ER -  }}';
       $this->assertEquals('ZooKeys', $expanded->get('journal'));
       $this->assertEquals('445', $expanded->get('issue'));
       $this->assertNull($expanded->get('volume'));
+      sleep(1);
       $text = '{{Cite journal|doi=10.3897/zookeys.445.7778|journal=[[Zookeys]]}}';
       $expanded = $this->process_citation($text);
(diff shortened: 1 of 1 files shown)


tests/phpunit/TemplateTest.php
```diff
@@ -41,8 +41,8 @@ protected function process_page($text) {  // Only used if more than just a citat
 
   public function testDisableParallelBuilds() { // This test should be first one in file
      if (PHP_VERSION_ID < 60000) {  // Either 5.* or 7.*
-         echo "\n Sleeping PHP 6 build for 60 seconds \n";
-         sleep(60);  // Delay the non-7 build for 30 seconds to keep us from hitting the API wall
+         echo "\n Sleeping PHP 6 build for 120 seconds \n";
```
ms609: If APIs have such a large window, then they don't sound scaleable: are they really designed to be used in a bot context?

GlazerMann: I have been avoiding it, but I can get 10 Keys easily and will wrote code to rotate through them.

I have been waiting for google to up my quota on this key

GlazerMann: Array of ten keys by the end of the weekend :-(

PR #378, opened 2017-12-04
Prefer Books.Google.Com to GoogleAPIs.Com

GlazerMann:
Some research shows that:
1.  the Google Books API has a "You must use an API key" license.  This is because the API is a very rich method for polling the Google Database.  Also, it has the ability to maintain personal libraries, interface with bookstores, etc..
2.  Books.Google.Com has a very liberal license.  
3.  We already use Books.Google.Com for {{citation|url=books.google.......+}}
4.  For ISBNs, we can get away with Books.Google.com
5.  Still need API for LCCN and OLCN
6.  I have applied for a bigger quota, but have not heard back

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1364,20 +1364,20 @@ final class Template {
         if ( !ctype_alnum($oclc) ) $oclc='' ;
       }
       if ($isbn) {  // Try Books.Google.Com
-          $google_book_url='https://books.google.com/books?isbn='.$isbn;
-          $google_content = @file_get_contents($google_book_url);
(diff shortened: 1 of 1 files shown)


ms609:
I do have recollections of trying with the Google API back in the day, and giving up, though I can't remember why.  If we are using Google's data then we really ought to be using their API to access it - otherwise I imagine that we'll be in breach of their terms of service.  If the API doesn't allow us enough volume to be useful, then my view would be that we ought to respect the limits that Google have set.

PR #383, opened 2017-12-05
Respect Page() protected members

GlazerMann:
The three APIs used $page->stuff directly.
This is all my fault for adding "protected" to those variables without looking in the php binaries


GlazerMann:
This is all my fault  for adding "protected" to those variables without looking in the php binaries

PR #391, opened 2017-12-10
Functions now in classes

GlazerMann:
Some funtions got moved into classes,  Need $Class->


ms609:
Hope I resolved the conflict correctly!

PR #392, opened 2017-12-10
Better user output

GlazerMann:
A likelihood of 1/12 looks bad.  Change scale to 13-24 from 1-12.  Purely to make users not scratch their head when the see us doing stuff with a likelihood of one out of twelve.


ms609:
Presumably the other metric ought to be adulterated in the same way?

GlazerMann:
the old 1-12 is now 13-24

the old 7-12 is now 14-24

PR #395, opened 2017-12-12
Move query adsabs to class where called

GlazerMann:

### DIFF
--- a/DOItools.php
+++ b/DOItools.php
@@ -4,7 +4,6 @@ require_once('WikipediaBot.php');
 /* junior_test - tests a name for a Junior appellation
  *  Input: $name - the name to be tested
  * Output: array ($name without Jr, if $name ends in Jr, Jr)
- * @codeCovIgnore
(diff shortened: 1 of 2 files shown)


ms609:
Knowing very little about programming good practice, when is it preferable to include a function within a class rather than the global namespace?

GlazerMann:
That is a good question.   In some languages and with some compilers it can be a performance boost.   That is not the case here.   I am mostly self taught too.  In this case, it is because we only call it from there and we do the same with books, etc.   it is just a consistent style issue.

GlazerMann:
Some codes have every function in a file.  Another code just had everything in single 27,000 fortran subroutine.   That’s wrong on so many levels.   It’s style only.   Since that function is only called from within that class, it “makes sense” to have it there; but, I am “software engineer” in job title only; having only taken one programming class before fortran 90 even existed.  That was back when I thought implicit in fortran was a good idea.

PR #399, opened 2017-12-16
Go directly to jstor RIS, skip citoid

GlazerMann:
Also fixed RIS if the page numbers are in the wrong order.
Please look at my skipping of ISSN, etc.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1200,7 +1200,6 @@ final class Template {
             case "EP":
               $end_page = trim($ris_part[1]);
               $dat = trim(str_replace("\n$ris_line", "", "\n$dat"));
-              $this->add_if_new("pages", $start_page . "-" . $end_page);
               break;
             case "DO":
               $ris_parameter = "doi";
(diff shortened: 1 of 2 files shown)


ms609:
Looks neat -- there was at one point a RIS data-extractor in the code - did this go, or is it worth using the same code for both purposes?

Template.php
```diff
@@ -1158,9 +1158,105 @@ protected function expand_by_doi($force = FALSE) {
     }
   }
   
+  protected function expand_by_jstor() {
+    if ($this->incomplete() === FALSE) return FALSE;
+    if ($this->blank('jstor')) return FALSE;
+    $jstor = $this->get('jstor');
+    if (preg_match("~[^0-9]~", $jstor) === 1) return FALSE ; // Only numbers in stable jstors
```
ms609: Will $jstor already be trimmed (I believe so?), or could we end up with whitespace at one end or the other?

Template.php
```diff
@@ -1713,6 +1809,13 @@ protected function use_unnamed_params() {
             $dat = trim(str_replace("\n$ris_line", "", "\n$dat"));
           }
         }
+        if (isset($start_page)) { // Sometimes people put end page before start page in reference
+          if (isset($end_page)) {
+             $this->add_if_new("pages", $start_page . "-" . $end_page);
```
ms609: It looks in this editor like a hyphen between start and end page – should be an en-dash.

Template.php
GlazerMann: You are right.  Some moms where I live use Vic’s vapor rub for all problems.  I similarly apply trim() to all texts.  I will remove

GlazerMann:
I copied RIS code.   I should make it into a function and thus have only one copy.

GlazerMann:
I believe this is now done

PR #402, opened 2017-12-18
http to https

GlazerMann:
Fixes http usage that will probably be turned off some day

AdsAbs already did it


GlazerMann:
Pleasantly surprised that they all just worked

ms609:
@GlazerMann: Could you take a look at the conflicts and either resolve these by updating the patch, or advise on how they ought to be resolved?

PR #408, opened 2018-01-26
add test for funky characters

GlazerMann:
Just a test


GlazerMann:
I will update once it runs

PR #413, opened 2018-02-26
Fix earlier pull on page ranges

GlazerMann:
expanded pages=123 to pages=123-123  
String comparisons are hard.


GlazerMann:
The failure is JSTOR being not returning anything one SICI query

PR #418, opened 2018-07-09
debug query_adsabs

GlazerMann:
Maybe the key is invalid/expired???  We are getting nothing back from the database


ms609:
Oddly, adsabs works fine on my local machine.

GlazerMann:
opening a different pull

PR #432, opened 2018-07-24
only forget publisher for journals if it is same

GlazerMann:
Includes a test
I believe this is ready to merge 
I am NOT saying that this is a good idea.   It is still being discussed

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -108,12 +108,12 @@ final class Template {
           $this->name = 'Cite journal';
           $this->rename('eprint', 'arxiv');
           $this->forget('class');
-          $this->forget('publisher');
+          if ($this->has('publisher') && $this->has('journal') && strcasecmp($this->get('publisher'),$this->get('journal')) === 0) $this->forget('publisher');
         } else if ($this->has('doi')) { // cite arxiv does not support DOI's
           $this->name = 'Cite journal';
           $this->rename('eprint', 'arxiv');
           // $this->forget('class');      Leave this for now since no journal title
-          $this->forget('publisher');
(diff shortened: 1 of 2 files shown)


josve05a:
This is a parameter field which adds content and should not be removed (by a bot) without consensus, not the other way around. We should not get consensus in order to NOT remove the parameter.

A manually entered field (which add content) should not be removed by a bot. A manually entered field which is redundant or does not add anything new could be removed by a bot. To change this, consensus needs to be gathered, not the other way around.

GlazerMann:
https://en.wikipedia.org/wiki/Wikipedia:Citing_sources   Only suggested for books.

https://en.wikipedia.org/wiki/Help:Citation_Style_1#Work_and_publisher  should not be included unless important 

This feature is over three years old.

GlazerMann:
The bot has been removing it since February 2014 which is when the code went to GitHub.  And it was certainly not new at that point

josve05a:
Lack of complaint does not equal consensus. Has the bot (previously, or ever) gotten community consensus to remove such fields?

ms609:
From memory, Cite journal used to specify that publisher should not be set for journals; as such, the bot was implementing a consensus decision.  I note now that the documentation does not explicitly mention this.  I'm not aware of any style guide that would advocate including a journal's publisher, as it is very difficult to think of a situation where this would be useful information.  Nevertheless, I would suggest that Template talk:Cite journal would be an appropriate place to reach a consensus on the appropriate behaviour here.

I'm closing the PR for now; feel free to reopen it once consensus has been obtained / reaffirmed.

PR #435, opened 2018-07-24
Avoid Series and Journal names that are the same

GlazerMann:
Multiple databases sometimes define series and journal title differently
Includes a test now


ms609:
Do we need to add or amend any tests in relation to this change?

GlazerMann:
Looks good

GlazerMann:
@ms609 test added

PR #438, opened 2018-07-25
Do not add Google Books URL-part 2

GlazerMann:
Includes a test to make sure we do not add urls


GlazerMann:
@ms609 I think this is good

PR #448, opened 2018-07-28
add one more journal

GlazerMann:


ms609:
Should there be a space before the "Historical"?  
More generally, is there any merit in implementing the rule "always capitalize the letter after a colon"?  This might help to reduce the number of manual exceptions that we must specify.

GlazerMann:
Fixed.  You are right about needing a more general fix.  Merge this and we can do general fix soon

ms609:
Currently failing constantsTest::testConstantsDefined:
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-' Historical Biology: An International Journal of Paleobiology '
+' Historical Biology: An International Journal Of Paleobiology '

GlazerMann:
It now works.
Another more general pull exists too now

PR #472, opened 2018-07-31
AdsAbs header reads

ms609:


ms609:
Better! But more to come...

PR #477, opened 2018-08-01
Stop curl from printing output

GlazerMann:
CURL is printing standard out to log files 
This information is all present in the CURL data we parse, including errors etc.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -654,7 +654,7 @@ final class Template {
         $ch = curl_init($encoded_url);
         curl_setopt($ch, CURLOPT_HEADER, 1);
         curl_setopt($ch, CURLOPT_NOBODY, 1);
-        curl_setopt($ch, CURLOPT_VERBOSE, 0); // stop curl from printing non-errors to stdout
+        curl_setopt($ch, CURLOPT_VERBOSE, 0); // stop curl from printing everything to standard out
         if (curl_exec($ch)) {
           $redirect_url = curl_getinfo($ch, CURLINFO_REDIRECT_URL);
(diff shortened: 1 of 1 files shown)


Template.php
```diff
@@ -2463,6 +2464,7 @@ function assessUrl($url){
         curl_setopt($ch, CURLOPT_NOBODY, 1);
         curl_setopt($ch, CURLOPT_HEADER, FALSE);
         curl_exec($ch);
+        curl_setopt($ch, CURLOPT_VERBOSE, 0); // stop curl from printing non-errors to standard out
```
ms609: Presumably this ought to come before the `curl_exec` on the previous line?

Template.php
```diff
@@ -654,6 +654,7 @@ protected function get_identifiers_from_url($url_sent = NULL) {
         $ch = curl_init($encoded_url);
         curl_setopt($ch, CURLOPT_HEADER, 1);
         curl_setopt($ch, CURLOPT_NOBODY, 1);
+        curl_setopt($ch, CURLOPT_VERBOSE, 0); // stop curl from printing non-errors to stdout
```
ms609: Ought we to (or can we) set this option globally rather than repeatedly?

Template.php
GlazerMann: Ooops.  Fixed.

Template.php
GlazerMann: I do not know.  It is only actually needed here, since this is the only place in the code that we have curl return TRUE/FALSE instead of an object.

PR #484, opened 2018-08-01
add comments to defend bot's actions

GlazerMann:
I do not want these lines ever removed.
Even if we stop removing publisher from cite journal at some point down the road, these two removals need to still happen.


ms609:
I don't quite understand what these comments mean, I'm afraid..?

GlazerMann:
There are three places we remove publisher
If concenious changes - I hope it never does - we still need the two removals

PR #487, opened 2018-08-01
Stop curl from writing to stdout

GlazerMann:
previous fix did not work


ms609:
I'm not entirely clear where you are seeing cURL write to stdout?

ms609:
Gotcha, ta.  My first suspicion would be that I've left a debugging line in the code somewhere (`echo $ch` or something).  I'll take a look tonight and see whether it's easy to clear up.

GlazerMann:
Basically CURL always sends output to standard out, unless you tell it otherwise.

PR #498, opened 2018-08-03
I think YAML wants ==

GlazerMann:


ms609:
Copied this from some docs somewhere, with the single equals... I'll merge the request and we can watch what happens!

PR #507, opened 2018-08-05
Forget "work=" if same as other text

GlazerMann:
We never set work, but people do.
Also do the template clean-up for more than just cite journal


ms609:
Can we add a test for this new behaviour, please?

PR #514, opened 2018-08-06
Do not add OA, if title has a link

GlazerMann:


ms609:
Should there ever be links in the title field?  Sounds like this is the error (it will poison the metadata produced by the template, for example), and the bot should remove them if found.

PR #519, opened 2018-08-07
Do not add bad DOI

GlazerMann:


ms609:
Some DOIs (Yes Jstor, I'm thinking of you) are correct, but are not resolved by CrossRef.  Should such DOIs not still be added?

GlazerMann:
Research Gate too.

PR #524, opened 2018-08-09
Deal with URLs as Journal names

GlazerMann:
This fixes us from making garbage input even worse.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -2441,9 +2441,11 @@ final class Template {
             $this->forget('publisher');
           case 'periodical': 
             if (substr(strtolower($p->val), 0, 7) === 'http://' || substr(strtolower($p->val), 0, 8) === 'https://') {
-              if ($this->blank('url')) $this->rename($pmatch[1], 'url');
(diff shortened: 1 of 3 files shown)


ms609:
Two musings:
- Are the test cases legitimate input that ought to be supported by Cite journal etc?  Seems to me like this behaviour might be caught under GIGO.
- This seems like a heavy-handed approach.  Better (if perhaps unduly complicated?) would be to support URLs in `title_capitalization` -- i.e. to include a regexp in the function that allows parts of the string that SHOULD be capitalized to be so. 
e.g. `[http://journalofmedicine.com journal of medicine]` → `[http://journalofmedicine.com *J*ournal of *M*edicine]`

GlazerMann:
It is really GIGO.  We just do not want to make it worse.
I have never actually seen the [htt....] case.

GlazerMann:
I think I need to change is so that if we find www... then we rename publisher to website and if we find http.... we rename it to url

GlazerMann:
@ms609 How about this?

PR #525, opened 2018-08-09
Remove Wikilinks from journal & titles names

GlazerMann:
The only thing we leave is if the entire name of a journal is a wikilink


GlazerMann:
@ms609 It works.  Your wikilink examples were evil.

PR #539, opened 2018-08-10
French d'Words and l'Phrases capitalization

GlazerMann:


expandFns.php
```diff
@@ -221,6 +221,17 @@ function title_capitalization($in, $caps_after_punctuation) {
     create_function('$matches', 'return mb_strtolower($matches[0]);'),
     trim($new_case)
   );
+  
+  $new_case = preg_replace_callback("/ (L')([a-zÀ-ÿ]+)/", function($matches) { /** French l'Words  **/
```
ms609: Do we also need to match unicode alternatives to the apostrophe, i.e. &rsquo; and kin?

expandFns.php
```diff
@@ -221,6 +221,17 @@ function title_capitalization($in, $caps_after_punctuation) {
...
+    return " l'" . mb_ucfirst($matches[2]);
+  }, ' ' . $new_case);
+  $new_case = mb_ucfirst(trim($new_case));
+
+  $new_case = preg_replace_callback("/ (D')([a-zÀ-ÿ]+)/", function($matches) { /** French d'Words  **/
```
ms609: You can use \b to match a word boundary, i.e. `\b(D').  And there's no need for brackets around the (D') / (L') as if there is a match, we don't need to refer to $matches to figure out what it will be!

GlazerMann:
@ms609 Thoughts?  I made your tests even harder.

GlazerMann:
@ms609 It actually works now

expandFns.php
```diff
@@ -221,6 +221,20 @@ function title_capitalization($in, $caps_after_punctuation) {
     create_function('$matches', 'return mb_strtolower($matches[0]);'),
     trim($new_case)
   );
+  /** French l'Words and d'Words  **/
+  $new_case = preg_replace_callback(
```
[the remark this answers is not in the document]
GlazerMann: Yes all

PR #541, opened 2018-08-11
7.1.11 are we ready?

GlazerMann:
running tests with 7.1.11 to see if we are using features that will go away at some point


GlazerMann:
Turned off 5 for speed.
Never intended to really commit from this branch.  Just testing.

PR #547, opened 2018-08-11
Try to use open URL access even if we already have url

GlazerMann:
Might get PMC, ARXIV, HDL, etc.


ms609:
Is it possible to add a test case that verifies that this works?

GlazerMann:
Test added

PR #550, opened 2018-08-12
Convert "work" if possible to better choice

GlazerMann:


Template.php
```diff
@@ -207,6 +207,15 @@ public function process() {
         if ($this->has('chapter') && ($this->wikiname() == 'cite journal') && ($this->get('chapter') != $this->get('title') || $this->has('isbn'))) { 
           $this->name = 'Cite book';
         }
+        if ($this->wikiname() === 'cite journal' && $this->has('work') && $this->blank('journal')) { // Never did get a journal name....
+          $this->rename('work','journal');
```
ms609: Would you mind just checking the whitespace after commas before I merge? 
Cheers!

GlazerMann: I will do that.
We really should add a php_lint step to the rest suites.  I am only half joking.

PR #553, opened 2018-08-13
Fix bugs that crash bot

josve05a:


GlazerMann:
I will create a non-conflict pull

GlazerMann:
High-jacked pull to make emergency fix

PR #565, opened 2018-08-15
More Small words and more testing 

GlazerMann:
Test constants for exactly one leading and trailing spaces.

U S A never worked either; fixed that.

Special code for "Ann. Of " to lose the capital O.

One giant test for all words added too.

Add some more words


ms609:
This is a very good commit! (-:

PR #576, opened 2018-08-16
ready for https in bibcode URLS

GlazerMann:
They might someday do that.  Also, someone (or some bot) might change it to https and then we would not recognize it.  Future-proofing the code.


GlazerMann:
Fixed.  I should note the we have grant agencies demanding https be available, and some warning us that they plan to at some point refuse to fund anyone that allows http, so this is only a matter of time.

PR #577, opened 2018-08-16
Use https

GlazerMann:
Plus, the w3.org ones all redirect to https anyway. And......

Since web browsers are implemented with special-purpose HTML parsers, rather than general-purpose DTD-based parsers, they don't use DTDs and will never access them even if a URL is provided. The DOCTYPE is retained in HTML5 as a "mostly useless, but required" header only to trigger "standards mode" in common browsers


ms609:
A long time since I copied and pasted this template from somewhere, so feel free to streamline it!

PR #596, opened 2018-08-17
Format HTML output in category

GlazerMann:
The PHP lint option actually pointed out that my first submit was rubbish.....


ms609:
Could the calls to `html_echo` be reduced by putting the output in a `<pre>` wrapper as in DoiBot.php, to cause `\n` to be rendered in HTML?

PR #599, opened 2018-08-17
Test PARAMETER_LIST against wikipedia

GlazerMann:
Done and ready to commit:
Looks for stuff we have that is extra--not an error, since we have {{cite patent}} stuff on our list.
Looks for stuff wikipedia has that we do not have
Looks for duplicates in our list
Runs sort() on the list and prints out any errors


constants/parameters.php
```diff
@@ -272,4 +272,14 @@
  'translator-last', 'translator-last##','translator-first', 'translator-first##',
  'translator##-last', 'translator##-first',
  'interviewer##', 'interviewer##-first', 'interviewer##-last', 'interviewer##-link', 'interviewer##-mask', 'interviewer-first##', 'interviewer-last##', 'interviewer-link##', 'interviewer-mask##', 
- );
+ 'call-sign', 'chapter-link', 'day', 'publisherid', 'titleyear', 'air-date', 'book-title', 'chapter-url-access', 'contributor',
+ 'contributor-first', 'contributor-last', 'contributor-given', 'contributor-surname', 'contributor-link', 'contributor-mask',
```
ms609: We should use alphabetical order to make future maintenance easier.  We could even add a test that checks for alphabetical order?

GlazerMann:
@ms609 let’s just accept this and I will do alphabetical later

PR #601, opened 2018-08-18
Only add class= to arxiv

GlazerMann:
class is not a very useful parameter
consensus not needed for this change since we just add it less often -- not removing it.


GlazerMann:
No longer removes parameter.  Just adds it less often.

GlazerMann:
https://en.wikipedia.org/wiki/Help_talk:Citation_Style_1#Remove_|class=_from_non-cite_arxiv_templates

PR #604, opened 2018-08-19
Remove outer quotes from titles

GlazerMann:


ms609:
Worth catching single quotes whilst you are at it?

PR #606, opened 2018-08-19
Better formatting of category output on the web

GlazerMann:


category.php
```diff
@@ -2,6 +2,7 @@
 <?php
 
 error_reporting(E_ALL^E_NOTICE);
+html_echo('<br><pre>','\n');
```
ms609: Will this `<pre>` ever be closed?

GlazerMann: Not really important but will do it

GlazerMann:
Just realized that HTML output was not getting set.  Fixed that.

PR #614, opened 2018-08-19
Deal with empty work= parameter better

GlazerMann:
No longer just delete it.  Change it to encourage better editing in the future.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -236,14 +236,27 @@ final class Template {
         }
       }
       // "Work is a troublesome parameter
-      if ($this->get('work') !== NULL) { // We want to catch {{Cite|work=}} also, so do not use $this->has('work')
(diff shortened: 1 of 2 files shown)


Template.php
```diff
@@ -236,14 +236,27 @@ public function process() {
...
+             $this->has('periodical') ||
+             $this->has('website')) {
+              $this->forget('work'); // Delete if we have alias
+         } elseif ($this->wikiname() === 'cite web') {
+            $this->rename('work', 'website');
+         } elseif ($this->wikiname() === 'cite journal') {
+            $this->rename('work', 'journal');
+         } elseif ($this->wikiname() === 'cite magazine') {
+            $this->rename('work', 'magazine');
+         }
```
ms609: Do we need to handle $this->wikiname() === 'citation'?  Perhaps leaving `work` stand is the correct behaviour in that case.

GlazerMann: That illdefined template gives us no hint on what is better

PR #617, opened 2018-08-20
Do google books for cite web

GlazerMann:
Long standing problem

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -90,7 +90,7 @@ final class Template {
         $this->use_unnamed_params();
         $this->get_identifiers_from_url();
         $this->tidy();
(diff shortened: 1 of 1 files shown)


Template.php
```diff
@@ -90,6 +90,10 @@ public function process() {
         $this->use_unnamed_params();
         $this->get_identifiers_from_url();
         $this->tidy();
+        if (strpos($this->get('url') , 'google.com') && $this->expand_by_google_books()) {
```
ms609: Can we be more precise in our URL search, i.e. looking for a component that shows us that we're at google books not google finance (etc)?

GlazerMann: Done.

GlazerMann:
@ms609 I think this is good To go

PR #620, opened 2018-08-20
Better handling of wikilinks in titles

GlazerMann:
Removes all [[ and ]] from titles: that is not new.
Full title wikilinks become title-link: that is not new.
What's new:
   Single partial wikilinks get converted to title-link
https://en.wikipedia.org/wiki/User_talk:Citation_bot#Do_not_remove_partial_wikilinks_in_titles
  In the case it is GIGO this makes it more likely for a human to fix, so it is still an improvement.

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -218,7 +218,7 @@ final class TemplateTest extends PHPUnit\Framework\TestCase {
     $this->assertEquals('Dark Lord of the Sith Pure Evil', $expanded->get('title'));
     $this->assertEquals('Sith (Star Wars)', $expanded->get('title-link'));
   }
-      
+  
   public function testJournalCapitalization() {
     $expanded = $this->process_citation("{{Cite journal|pmid=9858585}}");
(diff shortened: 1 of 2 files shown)


tests/phpunit/TemplateTest.php
```diff
@@ -214,9 +214,11 @@ public function testRemoveWikilinks() {
     $this->assertEquals('Pure Evil', $expanded->get('title-link'));
     $expanded = $this->process_citation("{{Cite journal|title=[[Dark]] Lord of the [[Sith (Star Wars)|Sith]] [[Pure Evil]]}}");
     $this->assertEquals('Dark Lord of the Sith Pure Evil', $expanded->get('title'));
-  
+    $expanded = $this->process_citation("{{Cite journal|title=Dark Lord of the [[Sith (Star Wars)|Sith]] Pure Evil}}");
+    $this->assertEquals('Dark Lord of the Sith Pure Evil', $expanded->get('title'));
+    $this->assertEquals('Sith (Star Wars)', $expanded->get('title-link'));
```
ms609: Is this really the expected behaviour?  The link refers to part of the title, not the full title, so the user is unlikely to expect the full title to link to the partial link.

PR #634, opened 2018-08-23
Check for doi-inline template in title

ms609:
I've created this PR in place of a TODO comment in the text, so that the TODO eventually gets done!
Feel free to have a stab at a solution.


GlazerMann:
@ms609 look good to you?

Issue #458, opened 2018-07-29
Template::check_url


ms609:
Function disabled: comment reads:
    /*  Disable; to re-enable, we should log possible 404s and check back later.
     * Also, dead-link notifications should be placed ''after'', not within, the template.

josve05a:
Probably not the right place to mention this, but https://github.com/wikimedia/DeadlinkChecker is a good repository/dependency to check if a link is dead.

ms609:
Handy, thanks!

GlazerMann:
@ms609 putting this in the code

GlazerMann:
Should we add dead url tags when we find url titles like “buy this domain now!”

GlazerMann:
#1606 Should we close this issue, after this documentation is added.  I really do not feel like beating my head against the mobs on this issue at this time.

GlazerMann:
I will close since this is all documented in the source code and at this point in time, there are other bots that do this.

PR #636, opened 2018-08-23
adding information from ISSUES as comments

GlazerMann:


GlazerMann:
Fixes  #458

ms609:
My feeling is that issues are more visible than comments, and thus more likely to be actioned when the appropriate rainy day comes along -- or indeed to attract attention from others in the open source community.  So I'd prefer to keep the TODOs there if that's okay.

PR #659, opened 2018-08-24
This only works for ms609

GlazerMann:
this code runs on the pull request but the script dies a 99 and cannot go on
Sleep so that the code has time for the update to patch the patch 
I am surprised that the code exists with 99 but somewhere somehow the code gets patched

### DIFF
--- a/maintain_files.sh
+++ b/maintain_files.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-[ -z "${GITHUB_PAT}" ] && exit 99
+[ -z "${GITHUB_PAT}" ] && exit 0
 git config --global user.email "martins@gmail.com"
 git config --global user.name "Martin Smith"
 if [ "${TRAVIS_PULL_REQUEST}" = false ]
--- a/maintain_files.sh
+++ b/maintain_files.sh
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


ms609:
Maybe that's why it's an exit 0 by default.  Does it work without the sleep?

GlazerMann:
I am not sure.   Sleep is paranoia

GlazerMann:
I am concerned the next test will run before the update

GlazerMann:
I saw this in the capitals pull.   “You” edited my pull

ms609:
If the 99 code occurs, then presumably no pull is going to occur, as the script is exiting with failure instead of using the GITHUB_PAT?

ms609:
The idea is that small words can be added and will be sorted alphabetically by Travis as part of the build sequence.  This makes duplication easier to spot, and saves lists getting out of order.
Of course, it may be the case that this 'innovation' ends up causing more trouble than it saves...

GlazerMann:
I cannot seem to set the GitHub Pat for me in this repository.
Also, the script insists on setting itself to be you so I don’t know of that matters anyway

GlazerMann:
Maybe the sleep is not needed if the commit changes the code and makes Travis run again

PR #663, opened 2018-08-24
forget dates before adding new ones

GlazerMann:


apiFunctions.php
```diff
@@ -151,6 +151,8 @@ function arxiv_api($ids, $templates) {
         $current_year = $this_template->get_without_comments_and_placeholders('year');
         if (!$current_year
         ||  (preg_match('~\d{4}~', $current_year) && $current_year < $match[2])) {
+          $this_template->forget('date');
+          $this_template->forget('year');
```
ms609: Can we be confident that the data we've obtained from arxiv will always be more reliable than anything that the user has entered?

GlazerMann: Not really.  I should note that the code right now just overwrites the user input data.  This just makes sure that the old data is removed first.

GlazerMann: Not really.  I should note that the code right now just overwrites the user input data.  This just makes sure that the old data is removed first.  I think the idea is that if we find "journal data", then that is an upgrade to the arxiv date.

ms609: Fair enough.  I wonder whether we can do better in respecting user-input data.  In particular, we should use rename instead of forget if we can, to avoid disrupting internal formatting.

PR #664, opened 2018-08-24
improve arXiv data usage 

GlazerMann:
Parsing now all in a single function.
All Regex's are tested
Warning code for new regex's being needed added


GlazerMann:
The problem with using regex is that it does not realize that pages can be 2011, it thinks that they are years.

GlazerMann:
Using a regex is dangerous with:
    <arxiv:journal_ref xmlns:arxiv="http://arxiv.org/schemas/atom">Phys. Rev. Lett. 117, 211101 (2016)</arxiv:journal_ref>

ms609:
Perhaps we could use the test-led development paradigm here, and build a battery of tests (perhaps in this PR) that encompasses all the formats that we're likely to come across.  We can then hack at the code until all these cases are satisfied.

GlazerMann:
Very much test lead needed

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0106015

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0101014

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0102014

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0103014

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0107014

GlazerMann:
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=1708.07261
JAP, 122, 104902 (2017)
view-source:https://export.arxiv.org/api/query?start=0&max_results=2000&id_list=nucl-ex/0109014
Phys.Rev.A66:022109,2002

GlazerMann:
Code needs to detect case and then process.
If fits none of the six known case then prints a warning

GlazerMann:
Thoughts on new code?
Also need to copy to other place where same code exists

GlazerMann:
And I need to double check Regex

GlazerMann:
This is now "test driven" to the extent that I found three formats by just downloading some arXiv data.  The test suite found two more formats.  The code prints information when run if it finds something new (It is an error when running in Travis).

GlazerMann:
Should we turn this into a function call or is the template version going to be gone soon?

ms609:
Good question!  Worth looking at any calls to the template version and seeing whether they can be replaced.  Unfortunately I'm back in a time-poor phase so it might be a little while until I can finish the job...

GlazerMann:
@ms609 I am time poor now too.  Please accept this now and we can work on code duplication later

GlazerMann:
I will add these
1606.02558
1208.3207
0905.1039
0809.0880

GlazerMann:
code coverage shows all RegEx getting used

PR #672, opened 2018-08-25
Forget Google publishers

ms609:


GlazerMann:
A quick search found these
publisher=Google News Archive
publisher=[[Google]] 
publisher=Google News
publisher=Google.com
publisher=[[Google News]]

ms609:
gottem, thanks!

PR #677, opened 2018-08-26
Drop user name from researchgate URLs

GlazerMann:
Reduced number of tests

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -1038,19 +1038,7 @@ ER -  }}';
     
   public function testTrimResearchGateETC() {
     $want = 'https://www.researchgate.net/publication/320041870';
-    $text = '{{cite journal|url=https://www.researchgate.net/publication/320041870}}';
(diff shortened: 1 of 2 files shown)


tests/phpunit/TemplateTest.php
```diff
@@ -1053,6 +1053,9 @@ public function testTrimResearchGateETC() {
     $text = '{{cite journal|url=http://researchgate.net/publication/320041870_EXTRA_STUFF_ON_END}}';
     $prepared = $this->prepare_citation($text);
     $this->assertEquals($want, $prepared->get('url'));
+    $text = '{{cite journal|url=https://www.researchgate.net/profile/hello_user-person/publication/320041870_EXTRA_STUFF_ON_EN}}';
```
ms609: Do we need both tests here, or can we simply replace the previous test with the new one?

GlazerMann: I want to test both just to make sure we don’t accidentally require a / character.   I almost did that error

PR #689, opened 2018-08-27
reduce printing of internal information

GlazerMann:
Don't report strtolower('CITATION_BOT_PLACEHOLDER_BARE_URL') as unrecognized parameter

Or worse accidentally fix it!!!

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -2289,7 +2289,7 @@ final class Template {
   foreach ($this->param as $p) {
     ++$i;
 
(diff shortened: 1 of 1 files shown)


GlazerMann:
@ms609 This looks good.  Please accept to clean up logs users see.

ms609:
Not had the chance yet to explore why this is necessary, as I've added this parameter to our list of supported parameters (which seems the less ad-hoc way to go).  Perhaps I need to update the code on the server.

GlazerMann:
That works👍🏻

PR #691, opened 2018-08-28
Strip off citeseerx ends like  &rep=rep1&type=pdf

GlazerMann:


GlazerMann:
Test added thankfully since the code was broken

PR #694, opened 2018-08-29
Set and reset time limits

GlazerMann:
Two minutes for the next step and 45 seconds to load a page
That’s php time only not wall clock

FIXES: we were setting time limits and not resetting them.


ms609:
I wonder whether there's a value in resetting the time limit each time we begin a new citation – so that long pages can complete without timeout (what was that bug report recently with a mammoth N00 citation page?)

GlazerMann:
This code now ups the time limit to 120 and resets it often.  You are right,  we might need reset more places.

GlazerMann:
I say we try 2 minutes an then see if that page times out still.  If it does then we have a bunch of work to do. If not, then me move along since these are not the droids we are looking for.

PR #696, opened 2018-08-30
fix maintain files and git pull

### LINKED ISSUE #673: maintain_files not pushing
See https://travis-ci.org/ms609/citation-bot/jobs/420424097

### LINKED ISSUE #687: Gitpull.php not running?
It does not seem to be automatically pulling

Dev pull fails with an error too

GlazerMann:
I think this fixes  #687 and fixes #673
But, you will need to fix the Git state on wikipedia first.
Doing the git pull before and after is a little overkill, but better safe than sorry

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,14 +26,16 @@ matrix:
       if: branch = master AND type != pull_request
       php: 5.6 # Avoid downloading and installing 5.5
       script:
-        - git pull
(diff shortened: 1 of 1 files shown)


ms609:
I've cleaned up the Git state on ToolForge.  Will look at the changes when I have time so I can understand what's going on under the hood.

PR #697, opened 2018-08-30
Blow away local changes first

GlazerMann:
there should not be any.  But just in case.

### DIFF
--- a/gitpull.php
+++ b/gitpull.php
@@ -1,4 +1,6 @@
 <?php
+$output = [];
+exec ("git fetch --all", $output);
 exec ("git reset --hard", $output);
 exec ("git pull", $output, $return_var);
 ?><pre>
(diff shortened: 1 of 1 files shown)


ms609:
Before I merge – is there any danger that this will delete untracked files from the repo, such as the `env.php` file that I use to set environment variables?  I've accidentally deleted this with an unexpectedly powerful `git` command on more occasions than I'd care to repeat...

GlazerMann:
It is not supposed to.   I read the documentation several times before submitting this.

GlazerMann:
https://stackoverflow.com/questions/1125968/how-do-i-force-git-pull-to-overwrite-local-files

Some people just have a single decrypt key set and then add encrypted keys to git.  That way they only have one secret.

PR #699, opened 2018-08-31
add journal special cases

GlazerMann:
just copy and pasted listed on my phone


GlazerMann:
@ms609 please pull it in now

PR #704, opened 2018-09-01
Do not drop .pdf urls

GlazerMann:
Includes &uid= on end

Fixes deleting open acces copies

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -738,8 +738,10 @@ final class Template {
     if ($doi = extract_doi($url)[1]) {
       if (is_null($url_sent)) {
         if (doi_active($doi)) {
-          report_forget("Recognized DOI in URL; dropping URL");
-          $this->forget('url');
+          if (mb_strpos(strtolower($url), ".pdf") === FALSE) {
+            report_forget("Recognized DOI in URL; dropping URL");
+            $this->forget('url');
+          }
         } else {
           $this->mark_inactive_doi($doi);
         }


ms609:
We should include a test case here with an unambiguously necessary URL, so that this does not get removed in future.  (I'm still keen to see an example that cannot simply be reached through the DOI link.)

GlazerMann:
Yeah, we should probably add an array of sites to drop: sciencedirect.com etc.  (basically drop all the publishers).

GlazerMann:
Massive list of urls to drop added, such as "sciencedirect.com"

Template.php
```diff
@@ -738,8 +738,10 @@ protected function get_identifiers_from_url($url_sent = NULL) {
     if ($doi = extract_doi($url)[1]) {
       if (is_null($url_sent)) {
         if (doi_active($doi)) {
-          report_forget("Recognized DOI in URL; dropping URL");
-          $this->forget('url');
+          if ((mb_strpos(strtolower($url), ".pdf") === FALSE) || (str_ireplace($url, '', PUBLISHER_WEBSITES) != $url)) {
```
ms609: I can't follow the syntax of str_ireplace here.  Can we include a testcase, so that it's clear that this works?

GlazerMann: I will add a test case.   The logic is very odd, but it works.   The logic is simple but obscure.   You look for the publisher websites and replace them.   Then you look and see if the url changed, you know you found one.

ms609: Should the code read `str_ireplace(PUBLISHER_WEBSITES, '', $url)`?

GlazerMann: Ooops.  All fixed now.  Test added too.  Closing this conversation.  Not saying it is fixed, but it should now be obvious if test fails.

GlazerMann:
I am shock to see a bug in my newly tested code............

tests/phpunit/TemplateTest.php
```diff
@@ -211,10 +211,16 @@ public function testTemplateRenaming() {
   }
   
   public function testDoiExpansion() {
-    $text = "{{Cite web | http://onlinelibrary.wiley.com/doi/10.1111/j.1475-4983.2012.01203.x/abstract}}";
+    $text = "{{Cite web | http://onlinelibrary.wiley.com/doi/10.1111/j.1475-4983.2012.01203.x/files.pdf}}";
     $prepared = $this->prepare_citation($text);
     $this->assertEquals('cite journal', $prepared->wikiname());
     $this->assertEquals('10.1111/j.1475-4983.2012.01203.x', $prepared->get('doi'));
+    $this->assertNull($prepared->get('url'));
+    $text = "{{Cite web | http://freecopies.com/doi/10.1111/j.1475-4983.2012.01203.x/file.pdf}}";
```
ms609: Again, a real URL would be useful here, so that we can be sure that we're actually fixing the problem as it exists in the real world (if indeed it does).

tests/phpunit/TemplateTest.php
```diff
@@ -211,10 +211,16 @@ public function testTemplateRenaming() {
   }
   
   public function testDoiExpansion() {
-    $text = "{{Cite web | http://onlinelibrary.wiley.com/doi/10.1111/j.1475-4983.2012.01203.x/abstract}}";
+    $text = "{{Cite web | http://onlinelibrary.wiley.com/doi/10.1111/j.1475-4983.2012.01203.x/files.pdf}}";
```
ms609: Is this a real example?  Don't wiley now use `http://onlinelibrary.wiley.com/doi/pdf/10.1111/j.1475-4983.2012.01203.x`?

GlazerMann:
I probably should use real urls

GlazerMann:
I made up the urls since I have yet to find a journal PDF url that ends in .pdf.  Thus, the second test does not occur.

GlazerMann:
Todo: better URLs

tests/phpunit/TemplateTest.php
ms609: I'm not seeing the value in testing something that doesn't exist.

ms609:
Could it be that a .pdf ending is enough to signify that a link is not from a publisher, after all?

GlazerMann:
I have improved the user output.  I agree that the publisher list is overkill, but we do know that publishers do have pdf files

GlazerMann:
@ms609 I think this is ready to go.

ms609:
I still think that without real URLs to use as tests, there's no proven value behind this additional code complexity.

GlazerMann:
complexity removed @ms609

ms609:
> Could it be that a .pdf ending is enough to signify that a link is not from a publisher, after all?

Against this assertion, here's a real life publisher link ending in .pdf (but not containing a doi...) http://jcs.biologists.org/content/joces/s3-100/49/89.full.pdf

GlazerMann:
I guess we just accept this pull as it is, and wait for an example of a URL that:
1. contains DOI
2. ends in PDF
3. is a publisher's website
Once we find one of those, start forgetting those urls

GlazerMann:
prepare_citation() did the trick

PR #706, opened 2018-09-01
Revert "better handling of ends of titles."

ms609:
Reverts ms609/citation-bot#703

On reflection, this will delete closing semicolons from HTML entites; e.g. `title=example&#153;`

Perhaps we should just leave any semicolons for now?

(Should probably create this issue in a better way but time is short!)


GlazerMann:
See #707

PR #719, opened 2018-09-03
Website (like work) is alias for journal

GlazerMann:


ms609:
Can we add a test please?  The hope is that PRs ought in general to pass the codecov/patch check (though the codecov/project check will often report a fail due to the tests that are necessarily skipped in pull requests).

GlazerMann:
Test added

PR #724, opened 2018-09-05
Doi's with square brackets cannot search PUBMED

GlazerMann:


ms609:
Should a similar change be made in apiFunctions.php, to avoid passing [] dois to the batch pubmed search?

GlazerMann:
I do not think that the current code in apiFunctions.php searches based upon doi.  I might be wrong

ms609:
Ah yes, that's right.  Merged.

PR #726, opened 2018-09-06
another arxiv test

GlazerMann:


GlazerMann:
This test is because this was failing on Wikipedia.  Turns out the cause is that the citation bot is currently pulling the dev branch and not master.

PR #732, opened 2018-09-06
Avoid access violation

GlazerMann:
The current code crashes


GlazerMann:
It does force it, but it does not unset via= etc.   I don’t think forget has any checks.  Also the unset code does crash since all the param stuff is private/protected

GlazerMann:
@ms609 done

PR #734, opened 2018-09-06
Ready for PHP 7

GlazerMann:
Fixed blindly continuing after fatal error


GlazerMann:
@ms609 Better.  Thank you.

PR #736, opened 2018-09-07
Do not search book bibcodes

GlazerMann:
Need to learn to parse this data before we use it.
http://adsabs.harvard.edu/abs/2003hoe..book.....K
Add bogus journal data, etc.


GlazerMann:
@ms609 tests added

PR #741, opened 2018-09-09
Extract oclc parameters from urls

GlazerMann:


GlazerMann:
I wonder if the basis of the code coverage has shifted over time.  We no longer cover some code since we no longer look for square bracket containing DOIs on pubmed.

PR #742, opened 2018-09-09
Citoid API generalization

GlazerMann:
Not using code, but update to request and URL not jstor specific


GlazerMann:
Not sure about limitations.  We hit some that made me stop using this code for jstor.

ms609:
Fair enough.  When I looked into citoid, it looked like the answer was going to be to run our own instance of the API on our toolforge server.  Got part way into setting this up when life got started again.  I'll merge this in for now, but this function probably has a short life expectancy, particularly if it is not currently used.

PR #747, opened 2018-09-09
New page ranges should use long dashes

GlazerMann:
Fixes having to run bot twice


ms609:
This fix does the job, but it feels like it shouldn't be necessary: do you know why this wasn't caught by the final call to `tidy()`?

PR #759, opened 2018-09-11
Support REGEXP_EN_DASH too

GlazerMann:
Yet another dash to recognize in pages


Template.php
```diff
@@ -558,7 +558,8 @@ public function add_if_new($param_name, $value, $api = NULL) {
       case "page": case "pages":
         if (( $this->blank("pages") && $this->blank("page") && $this->blank("pp")  && $this->blank("p") && $this->blank('at'))
                 || strpos(strtolower($this->get('pages') . $this->get('page')), 'no') !== FALSE
-                || (strpos($value, chr(2013)) || (strpos($value, '-'))
+                || (strpos($value, chr(2013)) || (strpos($value, '-')) || (strpos($value, REGEXP_EN_DASH))
```
ms609: Is `REGEXP_EN_DASH` compatible with `strpos`?

tests/phpunit/TemplateTest.php
```diff
@@ -864,7 +864,7 @@ public function testJstorSICI() {
    $this->assertEquals('1961', $expanded->get('year'));
    $this->assertEquals('81', $expanded->get('volume'));
    $this->assertEquals('1', $expanded->get('issue'));
-   $this->assertEquals('43', $expanded->get('pages'));
+   $this->assertEquals('43–52', $expanded->get('pages'));
```
ms609: This test is only designed to check that `43` is extracted from the SICI; if the metadata is found from other sources then the test no longer tests what it is meant to test (and thus needs re-thinking...) – perhaps `process_citation` needs replacing with a function that has a narrower scope.

GlazerMann: new test added

Template.php
GlazerMann: Shocking yes.  I actually had to test it out.  Good question

PR #762, opened 2018-09-12
Mess with work less

GlazerMann:
This non-change pisses people off a shockingly amount.
Secondly it is GIGO often since bad information seems to put in the work parameter that should have been put somewhere else


ms609:
Some aspects of bot improvement open fascinating windows onto the human psyche.  
Seldom else does so little matter so much to so many...

PR #771, opened 2018-09-13
Do not add Accepted/submitted manuscript in |format=

josve05a:
This is both the wrong parameter ("type" would be a better one) and is not recomended by wither template or MoS style-guideline.
https://en.wikipedia.org/wiki/User_talk:Citation_bot#Misuse_of_Format

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1737,7 +1737,7 @@ final class Template {
           switch ($best_location->version) {
             case 'acceptedVersion': $type = 'Accepted manuscript'; break;
             case 'submittedVersion': $type = 'Submitted manuscript'; break;
-            //  case 'publishedVersion': $type = 'Full text'; break; // This is the assumed default
+            // case 'publishedVersion': $type = 'Full text'; break; // This is the assumed default
             default: $type = NULL;
           }
           if ($type) $this->add('type', $type);
--- a/Template.php
+++ b/Template.php
@@ -1735,9 +1735,9 @@ final class Template {
             return FALSE;
           }
           switch ($best_location->version) {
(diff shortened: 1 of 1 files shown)


josve05a:
> type: Provides additional information about the media type of the source; format in sentence case. Displays in parentheses following the title. Examples: Thesis, Booklet, CD liner, Press release. Alias: medium.

> format: Format of the work referred to by url; for example: PDF, DOC, or XLS; displayed in parentheses after title. (For media format, use type.) HTML is implied and should not be specified. Automatically added when a PDF icon is displayed. Does not change the external link icon. Note: External link icons do not include alt text; thus, they do not add format information for the visually impaired.

I'm not sure if type is the correct parameter either (given the description of the parameter from the /doc page, that's why I simply commented out these rows of code until we've established if the bot should even be adding this to citation templates.

ms609:
I've been [[WP:BOLD]] and edited the `type` documentation so that it is less ambiguous.  Let's proceed on the basis that that parameter is the most suitable for this sort of information.

GlazerMann:
Can you add code in tidy_parameters in Template.php that will look at format and if it is any of the following rename it to format:
            case 'acceptedVersion': $type = 'Accepted manuscript'; break;
            case 'submittedVersion': $type = 'Submitted manuscript'; break;

And delete if this one
            // case 'publishedVersion': $type = 'Full text'; break; // This is the assumed default

That would undo that bot's previous work and change it

josve05a:
...yeah...no. I can't code for sh*t. I just wanted to help out and get this taken care of the best I could. Seems my abilities are way too limited.

GlazerMann:
I submitted a different pull request that cleans up old ones

GlazerMann:
Hijacking existing pull to get emergency work done

PR #783, opened 2018-09-16
[[file:img|12px|alt=words]] and such

GlazerMann:
Current code does [[]] and [[|]].  
Added up to this size:
[[File:Example.png|thumb|upright|alt=Example alt text|Example caption]]

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -75,23 +75,14 @@ final class Template {
   // Parts of each param: | [pre] [param] [eq] [value] [post]
   protected function split_params($text) {
     // Replace | characters that are inside template parameter/value pairs
-    $START = "~(\[\[[^\[\]\|]+)";
-    $STOP  =      "([^\[\]\|]+\]\])~";
-    $MID   =      "([^\[\]\|]+)";
-    $PIPE  = "\|";
-    $text = preg_replace($START . $PIPE . $STOP, "$1" . PIPE_PLACEHOLDER . "$2", $text);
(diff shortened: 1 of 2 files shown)


josve05a:
It might be "garbage in", but it is valid garbage which renders properly. The bot currently makes it worse.  And it is currently present on over 600 articles (insource:/title\=\[\[([Ff]ile|[Ii]mage)\:/ -- note that a few are not cite templaes though)

ms609:
By making it worse, the bot may help to draw attention to the incorrect information, increasing the chance that a human editor will step in and clean up the mess.

If it is a recurrent problem, then perhaps the cite templates can be modified to notice the bad input and notify users more prominently.  That way users will gain immediate feedback and be able to change their behaviour, rather than the bot stepping in months later to make an edit that the original user may not see.

josve05a:
If a "bodge" is incorrect but works, it should be "noted" with a CS1|2 error/maint. message rather than having the bot making the "bodge" both incorrect and not work. Rather ahve an error incorrect but "look nice" than incorrect and visually broken imo. There are more ways to draw attention to problems (such as maint. messages/categories/templates/etc.) than breaking a working template.

GlazerMann:
The example is real-world.

ms609:
Yes, I agree that a CS1/2 error message could be useful here, particularly given how widespread you have found the misuse to be!

GlazerMann:
@ms609 it is now done in a loop.

PR #789, opened 2018-09-18
Do not do GIGO, just drop it

GlazerMann:
This prevents people from filling it bogus parameters later


ms609:
This is a bold move.  Does it have consensus on WP?

PR #791, opened 2018-09-18
testMisspeltParameters

GlazerMann:


GlazerMann:
If there ever was a pull needing squash before pull.  This is one

constants/parameters.php
```diff
@@ -224,7 +224,18 @@
 const ISSUE_ALIASES = ['issue', 'number'];
 const PAGE_ALIASES = ['page', 'pages', 'pp', 'p', 'at'];
 
-
+// Just leave this code here for now
```
ms609: This comment should give a rationale for leaving the code, and a note specifying when the code should be removed (or it'll be there forever...)

PR #793, opened 2018-09-19
Fix Blank()

GlazerMann:
The code started doing arrays of arrays.  That does not work.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -290,12 +290,20 @@ final class Template {
     ));
   }
 
-  public function blank($param) {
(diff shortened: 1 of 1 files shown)


ms609:
Can you pinpoint to where this error (arrays of arrays) arises?  Might be neater to fix this at source.

GlazerMann:
if ($this->blank(["last2", COAUTHOR_ALIASES, "author"]))

and other such things

ms609:
Hmm, I did wonder whether that was legal.  I've replaced these instances with `array_merge`.  Let me know if you spot others that I've missed.

PR #798, opened 2018-09-19
reduce code duplication - adsabs

GlazerMann:
This bibcode code is better


GlazerMann:
Sounds like I need to merge the two subroutines and have it loop over singles or do big ones.   That would eliminate the duplicate and avoid to many bigs.  That is short sighted on their part

PR #801, opened 2018-09-20
Use Citoid API

ms609:
We now have an implementation running on the main server.  (Not the dev branch yet).

Travis testing uses the public API, for now.


ms609:
Having issues getting a stable implementation of the translation server at the moment; hopefully I'll get to the bottom of this with time!

GlazerMann:
I think that after we use Citoid we should call
$this->get_identifiers_from_url();

Also, we might someday decide to convert citexeerx and such to a temporary url etc

GlazerMann:
We also might want to add 
If !incomplete() then do not call Citoid

ms609:
> I think that after we use Citoid we should call
> $this->get_identifiers_from_url();
> 

That's already called in `prepare`, before we use Zotero, which doesn't change the URL

> If !incomplete() then do not call Citoid
Done (in `expand_by_zotero`).

PR #806, opened 2018-09-20
Handle DOI's with bad characters

GlazerMann:


apiFunctions.php
```diff
@@ -394,6 +394,7 @@ function expand_by_doi($template, $force = FALSE) {
 }
 
 function query_crossref($doi) {
+  $doi = str_replace(array("#", "<", ">"), array(urlencode("#"), urlencode("<"), urlencode(">")), $doi);
```
ms609: Could we generalize this logic by adding new constants to `constants.php`, following `DOT_ENCODE`?
Worth verifying that the other characters in DOT_ENCODE don't need urlencoding for crossref too.

PR #809, opened 2018-09-20
add some support for script-title

GlazerMann:


GlazerMann:
Your big fix applied.   I do not trust metadata to not have spaces in the xml/json/etc

PR #813, opened 2018-09-22
Better URL to DOI code

GlazerMann:
Handle better cases where DOI is already set. 
Handle both new=old and new != old
White space changes since the new code does the "add(doi)" before "forget(url)"; which is the correct order actually.  Old code could drop url and lose it, if an existing doi was set.  That can be a problem when the original doi is wrong or when the article has more than one doi (publisher, jstor, researchgate can name a few)


tests/phpunit/PageTest.php
```diff
@@ -87,7 +87,7 @@ public function testEmptyPage() {
...
+      $this->assertEquals(str_replace(" ","", "URL reference test 1 <ref name='bob'>{{Cite journal |doi = 10.1007/s12668-011-0022-5|title = Reoccurring Patterns in Hierarchical Protein Materials and Music: The Power of Analogies|journal = Bionanoscience|volume = 1|issue = 4|pages = 153–161|year = 2011|last1 = Giesa|first1 = Tristan|last2 = Spivak|first2 = David I.|last3 = Buehler|first3 = Markus J.|arxiv = 1111.5297}}< / ref>\n Second reference: \n<ref >{{Cite journal |pmc = 3705692|year = 2013|last1 = Mahajan|first1 = P. T.|title = Indian religious concepts on sexuality and marriage|journal = Indian Journal of Psychiatry|volume = 55|issue = Suppl 2|pages = S256–S262|last2 = Pimple|first2 = P.|last3 = Palsetia|first3 = D.|last4 = Dave|first4 = N.|last5 = De Sousa|first5 = A.|pmid = 23858264|doi = 10.4103/0019-5545.105547}}</ref> URL reference test 1"), str_replace(" ","", $page->parsed_text()));
```
ms609: What's all this `str_replace` business then.... we wouldn't be trying to avoid an issue with whitespace being unexpectedly modified, would we? (-;

GlazerMann: Not sure why the whitespace changes.  Any thoughts?

ms609: Hmm, it's odd that volume, issue etc should be modified as collateral damage... are they being "re-added" or overwritten when the new doi is added via `add_if_new`, perhaps?

Template.php
```diff
@@ -793,15 +793,26 @@ protected function get_identifiers_from_url($url_sent = NULL) {
...
-        if (doi_active($doi)) {
-          report_forget("Recognized DOI in URL; dropping URL");
+      if (strcasecmp($doi, $this->get('doi')) === 0) { // DOIs are case-insensitive
+        if (doi_active($doi) && is_null($url_sent)) {
+          report_forget("Recognized existing DOI in URL; dropping URL");
           $this->forget('url');
+        }
+        return FALSE;  // URL matched existing DOI, so we did not use it
+      }
+      if ($this->has('doi')) $this->mark_inactive_doi(); // Flag old one if bad
```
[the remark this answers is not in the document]
GlazerMann: That line assumes that the other pull gets accepted which allows bad doi's to be replaced with new ones.  So, it is checking the old one before trying to add the new one.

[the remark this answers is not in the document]
GlazerMann: removed that line.  probably too bold

GlazerMann:
@ms609 I think this is ready to go

PR #816, opened 2018-09-23
conference paper metadata uses "chapter-url"

monperrus:
- for conference papers, the `url` field is changed to `chapter-url`, because most open-access versions of conference papers point to the paper itself, and not to the whole proceedings

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -2287,14 +2287,14 @@ final class Template {
       }
     }
     if ($new_name === 'cite book') {
-            // all open-access versions of conference papers point to the paper itself
-            // not to the whole proceedings
-            // so we use chapter-url so that the template is well rendered afterwards
-            if ($this->blank('chapter-url')) {
-              $this->rename('url', 'chapter-url');
-            } elseif (0 === strcasecmp($this->get('chapter'), $this->get('url'))) {
-              $this->forget('url');
-            }  // otherwise they are differnt urls
+      // all open-access versions of conference papers point to the paper itself
+      // not to the whole proceedings
+      // so we use chapter-url so that the template is well rendered afterwards
+      if ($this->blank('chapter-url')) {
+        $this->rename('url', 'chapter-url');
+      } elseif (0 === strcasecmp($this->get('chapter-url'), $this->get('url'))) {
+        $this->forget('url');
+      }  // otherwise they are differnt urls
     }
   }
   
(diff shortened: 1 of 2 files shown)


monperrus:
Hi @ms609,

Thanks for your fast feedback.

> Perhaps the place to add this feature is under the function add_if_new, as it seems reasonable to assume that URLs that the bot adds to cite book or cite conference templates

I discovered the problem where a `cite journal` with the right open-access URL was changed to `cite book`, without changing `url` to `chapter-url`, yielding an incorrect link. For this problem, `add_if_new` does not seem to be solution.

> I can think of conference abstracts (admittedly not the same as papers) where a PDF of the abstract volume is available, but separate PDFs of individual abstracts are not.

This may happen. Let's call those two cases:

* case 1: the open access URL points to the full paper
* case 1: the open access URL points to the abstract

To my knowledge, most open-access URLs indexed in BASE and oadoi.org point to the full paper, and I would say that we have at least >95% of case 1) and the rest for case 2) (but I haven't made any systematic measurement).

WDYT?

GlazerMann:
I added some error checking to the code.  First attempt had an error itself. 🙄

monperrus:
@GlazerMann thanks a lot, it improves much over my naive version!

ms609:
Thanks for clarifying the problem.  Perhaps it is worth being more general and incorporating the new code within `change_name_to`?

monperrus:
> Thanks for clarifying the problem. Perhaps it is worth being more general and incorporating the new code within `change_name_to`?

That's a really good idea. @GlazerMann what do you think?

GlazerMann:
And code that does the opposite:  IF no url, no chapter, but a chapter-url then rename chapter url

GlazerMann:
We need a test.
One thought: if url comes before expansion then chapter= will not be set yet
In that case we could mindlessly set chapter-url and then count on final tidy to change it to url

GlazerMann:
Three separate changes.  If we are not sure about some, we can split into three.

GlazerMann:
@monperrus I know I have seen this happen, do you have some examples that we can add as a test?

monperrus:
Thanks @ms609, @GlazerMann a lot, that's great.

When will it be deployed in production?

ms609:
All pull requests should deploy automatically minutes after they are merged.

monperrus:
Continuous deployment for CitationBot! Awesome!

PR #819, opened 2018-09-24
use array ability of blank

GlazerMann:


GlazerMann:
It looks like lacks will return FALSE if the element is set to " ", but blank does a trim().  I all these cases, blank is better, I think.

GlazerMann:
I honestly cannot think of a reason that one would use lacks.

GlazerMann:
@ms609 Please look at the test pull that compares the two.

PR #823, opened 2018-09-25
forget publisher=Amazon when droppping book url

GlazerMann:


ms609:
Should amazon be added to "Bad publishers" in bad_data.php too?

GlazerMann:
@ms609 test added

PR #836, opened 2018-09-27
Bad data screening and more tests of Zotero

GlazerMann:


ms609:
I'd suggest using the model of 

    expand_by_zotero($expanded);

to keep the tests focussed on the zotero API (rather than CrossRef jumping in, etc).

GlazerMann:
I am stopping all but one test, just to reduce load.  Then start the others when the first one actually works.  I do not need to be told four times that I suck.

constants/bad_data.php
```diff
@@ -19,3 +19,5 @@
 const NON_JOURNAL_BIBCODES = array('arXiv', 'gr.qc', 'hep.ex', 'hep.lat', 'hep.ph', 'hep.th', 
                                    'math.ph', 'math', 'nucl.ex', 'nucl.th', 'physics');
 const NON_PUBLISHERS = array('books.google', 'google books', 'google news', 'google.co', 'amazon.com', 'archive.org'); // Google Inc is a valid publisher, however.
+const BAD_ZOTERO_TITLES = array ('on jstor', 'Browse publications', 'Central Authentication Service',
```
ms609: Can "on jstor" sometimes be appended to the end of an otherwise good title?  A better response there might be to strip the "on jstor" and keep the rest.

tests/phpunit/zoteroTest.php
```diff
@@ -56,19 +56,55 @@ protected function expand_via_zotero($text) {
     $expanded->tidy();
     return $expanded;
   }
-  
-  public function testZoteroExpansion() {
+
+  public function testZoteroExpansionRGJunk() {
+    $text = '{{Cite journal|url =  https://www.researchgate.net/publication/2344536}}';
```
ms609: The purpose of this test is not immediately apparent

tests/phpunit/zoteroTest.php
```diff
@@ -56,19 +56,55 @@ protected function expand_via_zotero($text) {
...
+  public function testZoteroExpansionPII() {
+    $text = '{{Cite journal|url = https://www.sciencedirect.com/science/article/pii/S0024379512004405}}';
+    $expanded = $this->expand_via_zotero($text);
+    $this->assertEquals('10.1016/j.laa.2012.05.036', $expanded->get('doi'));
+  }
+
+  public function testZoteroExpansionJstorBook() {
+    $text = '{{Cite journal|url=https://www.jstor.org/stable/j.ctt6wp6td.10?seq=9#metadata_info_tab_contents}}';
+    $expanded = $this->expand_via_zotero($text);
+    $this->assertEquals($expanded->parsed_text(), $text);
```
ms609: Not clear that this is testing a successful expansion

constants/bad_data.php
GlazerMann: If you get "on jstor" then the citoid stuff thought it was a website and almost everything is wrong.
Run it again later and you are good to go.

tests/phpunit/zoteroTest.php
GlazerMann: This is just a research gate page that gives rubbish title

PR #849, opened 2018-09-27
be even more paranoid about zotero

GlazerMann:


GlazerMann:
>  Hmm my excitement about Zotero is waning somewhat the more bad data I see...
I had a blind date like that years ago.

GlazerMann:
Perhaps we need a white list

GlazerMann:
@ms609 can you just accept this.  One test ran fine....

PR #850, opened 2018-09-27
Don't add authors to web pages

ms609:
Too likely to be garbage


ms609:
Of course if I needed a hobby, what I should really do is improve the [translators at Zotero](https://github.com/zotero/translators/)...

PR #852, opened 2018-09-27
Date check

ms609:
Fixes https://en.wikipedia.org/w/index.php?title=Mollusca&diff=prev&oldid=861484387


GlazerMann:
testPublisherRecognition seems to no longer work because of the DOI is more important than other stuff pull

GlazerMann:
" Bad data screening and more tests of Zotero" pull has a publisher test in it that still works

PR #858, opened 2018-09-28
Update capitalization.php

GlazerMann:
Washington DC.  Which got me thinking so I added AC also.  Current and the cooling.  NASA added for good luck too. DC D.C. AC A.C. NASA


josve05a:
How about adding dots between all two- or three-letter combinations if they are to be capitalized? Such as BMC, AI, JAMA, NYT etc...I've seen a lot of weird Gigo, and it's better to keep status quo than change to e.g. A.i, J.a.m.a, N.y.t. etc.

GlazerMann:
Ph.D.  B.Sc.  i.e. e.g.

GlazerMann:
@ms609 I created an issue to keep this idea in our minds

GlazerMann:
@ms609 can we accept this for now and the revisit with the ISSUE I created.  j.a.m.a. Does not become J.A.M.A.

PR #864, opened 2018-09-29
Expand more things

GlazerMann:
Add more jstor code and block zotero from doing jstor

Add code to convert researchgate urls to RIS request-still fails in the tools server.

Add code to to expand a bunch of stuff, including no crossref doi using zotero.
Also documented and commented out ones that did not work and why.

I had to look hard to find a doi that dx.doi.org and crossref both gave no meta data, but I did.

I have pulled parts of this out and made then different pulls


GlazerMann:
@ms609 Any idea why the jstor book URLs and the researchgate URLs  give internal errors?

I did add some code to make them directly grab the RIS data, but that did not seem to help.

PR #868, opened 2018-10-02
Ignore rubbish dates

GlazerMann:
Just a string of numbers with no spaces or punctuation is not a date.
Earlier we grab four numbers in a row


GlazerMann:
I will add check for year sanity in the $time area and then commit

PR #869, opened 2018-10-02
Create testBaseClass.php

GlazerMann:


GlazerMann:
Because it is included, the test code complains that it has no tests.  So, Travis “finds” this class no matter where you put it.

GlazerMann:
So, I could move it there, it would be for our benefit.

GlazerMann:
@ms609 Ready for your approval

PR #876, opened 2018-10-03
cite newspaper is just an alias, and is not paper

GlazerMann:
cite news seems to be the preferred template


GlazerMann:
@GlazerMann How's that look?

GlazerMann:
@GlazerMann can you accept this.
A website is down so two tests fail

PR #883, opened 2018-10-05
Fix bibcodes with & symbol with this API

GlazerMann:
Because of the protection the KEYS.  I do not think I can check any escaping of the &


GlazerMann:
Todo 
1.  Check escaping
2. Save it to ids

GlazerMann:
Actually we can, we just have to NOT urlencode them.  Ironically, the answer to encoding the special characters is to not encode the.

PR #887, opened 2018-10-05
Use dx.doi.org for DOIs

GlazerMann:

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -326,7 +326,19 @@ function expand_by_doi($template, $force = FALSE) {
   // run this function, so we don't check this first.
   
   $doi = $template->get_without_comments_and_placeholders('doi');
-  if (!$template->verify_doi()) return FALSE;
+  if (!$template->verify_doi()) { // Try non-crossref DOI resolver
+     $ch = curl_init();
+     curl_setup ($ch, 'https://doi.org/' . $doi);
+     curl_setopt($ch, CURLOPT_NOBODY, 1);
+     curl_setopt($ch, CURLOPT_HTTPHEADER, array("Accept: application/x-research-info-systems"));
+     curl_setopt($ch, CURLOPT_RETURNTRANSFER, TRUE);
+     $ris = @curl_exec($ch);
+     if ($ris == FALSE) return FALSE;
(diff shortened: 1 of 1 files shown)


Template.php
```diff
@@ -1496,6 +1496,17 @@ public function expand_by_RIS(&$dat) { // Pass by pointer to wipe this data when
     // Convert &#x__; to characters
     $ris = explode("\n", html_entity_decode($dat, NULL, 'UTF-8'));
     $ris_authors = 0;
+
```
ms609: Do we need a `foreach` here?  Could we not instead do a regular expression search along the lines of `preg_match('~\n\s*T[I1]\s+(.*)$', $match)`?

apiFunctions.php
```diff
@@ -325,8 +325,12 @@ function expand_by_doi($template, $force = FALSE) {
   // there will be few instances where it could not in principle be profitable to 
   // run this function, so we don't check this first.
   
-  $doi = $template->get_without_comments_and_placeholders('doi');
-  if (!$template->verify_doi()) return FALSE;
+  if (!$template->verify_doi()) { // Try non-crossref DOI resolver
```
ms609: Presumably we also need to modify `verify_doi` such that it recognizes valid non-CrossRef DOIs?  If so, we'll need a different test in this `if`.

GlazerMann: The real goal of verify doi is to to see if the crossref rich api is available or if we have to use doi.org.

GlazerMann: Maybe the name of the function needs changed

apiFunctions.php
```diff
@@ -416,6 +416,31 @@ function query_crossref($doi) {
...
+     } catch (Exception $e) {
+       $template->mark_inactive_doi($doi);
+       return FALSE;
+     }
+     if ($ris == FALSE || stripos($ris, 'DOI Not Found') !== FALSE || stripos($ris, 'DOI prefix') !== FALSE) {
+       $template->mark_inactive_doi($doi);
+       return FALSE;
+     }
+     $template->expand_by_RIS($ris);
+     report_action("Querying dx.doi.org: doi:" . doi_link($doi));
```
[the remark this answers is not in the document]
GlazerMann: Will do.  You are right.  Probably not a big deal, but always pays to be right.

PR #888, opened 2018-10-06
Prefer jstor to jstor doi which is often broken

GlazerMann:


Template.php
```diff
@@ -858,6 +858,10 @@ protected function get_identifiers_from_url($url_sent = NULL) {
         $this->set('url', $url); // Save it
       }
     }
+   
+    if (preg_match('~^(https?://[www.|]jstor.org/stable/)(10.2307/)(.*)$~', $url, $matches)) {
```
ms609: - [www.|] will match w, w, w, or anything.  Something close to  `(?:www\.)` is probably what you intend
- escape the `.` in jstor.org
- Why do we need to capture (10.2307/)?  Remove parentheses?
- `.+` should probably replace `.*` in final capture set; we want a non-empty JSTOR id.

ms609: Is there any value in adding a test, too?  Possibly a test will duplicate existing ones but at the least a test would have caught some of these.

GlazerMann: RegEx fixed.  Will add test

PR #894, opened 2018-10-06
Replace non-standard spaces with standard space

GlazerMann:
const array() not setup yet


tests/phpunit/TemplateTest.php
```diff
@@ -1412,6 +1412,16 @@ public function testBibcodeBook() {
       $expanded = $this->process_citation($text);
       $this->assertEquals($text, $expanded->parsed_text());
   }
+ 
+  public function testSpaces() {
+      $text     = "{{cite book|title=X X X X X X X X X X X X}}";
```
ms609: As it's not obvious to the eye, could you add a comment describing the nature of the spaces in `$text`, and a note within the test noting why this is the desired behaviour?

PR #898, opened 2018-10-07
"An Error Occurred Setting Your User Cookie"

josve05a:
Due to ["An Error Occurred Setting Your User Cookie"](https://en.wikipedia.org/w/index.php?search=insource%3A%2FAn+Error+Occurred+Setting+Your+User+Cookie%2F&title=Special%3ASearch&go=Go)


josve05a:
I'm going across and fixing these, but there were 60 articles with that specific term by the time of this pull. https://en.wikipedia.org/w/index.php?title=George_Soros&diff=prev&oldid=862960175

GlazerMann:
Travis CI seems to be sick right now

PR #904, opened 2018-10-09
Check RIS for bad data

GlazerMann:


tests/phpunit/TemplateTest.php
```diff
@@ -800,6 +800,25 @@ public function testRIS() {
      $this->assertEquals('Claude E.', $prepared->get('first1'));
      $this->assertEquals('379–423', $prepared->get('pages'));
      $this->assertEquals('27', $prepared->get('volume'));   
+   
+      $text = '{{Cite journal  | TY - JOUR
```
ms609: Could you add a comment to this test noting why we expect `NULL`, to future-proof the test?

PR #911, opened 2018-10-09
I broke the handle code

GlazerMann:


ms609:
Worth adding a test so it doesn't happen again?

PR #926, opened 2018-10-12
Reject � from zotero

GlazerMann:
If we we have that in the title, there is little hope

It seems that sometimes zotero sends us the Unicode replacement character instead.  Citoid on en.wikipedia.org does not have this problem.

Crossref gives us that sometimes instead of umlauts. We just take it from them since it is just one letter.

The bug report is titles that are 90% this character

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -71,7 +71,7 @@ class Page {
   public function parse_text($text) {
     $this->text = $text;
     $this->start_text = $this->text;
-    $this->modifications = array();
+    $this->modifications = array(); 
   }  
 
   public function parsed_text() {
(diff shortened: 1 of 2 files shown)


GlazerMann:
It seems that sometimes zotero sends us the Unicode replacement character instead.   Crossref gives us that sometimes instead of umlauts.  We just take it from them since it is just one letter.  The bug report is titles that are 90% this character

GlazerMann:
I think that figuring out zotero is the real solution; but we should probably reject anything with more than X of these or more than Y% of text being these things

PR #930, opened 2018-10-14
Deal with running out of big queries

GlazerMann:


GlazerMann:
No idea.   Just paranoid

PR #937, opened 2018-10-15
Make the “hard” a little harder & toss all changes

GlazerMann:
Do a full reset afterwards in case things failed too
I hope this will fix the problems.
I think this comes from multiple maintain files and git pulls going at once


GlazerMann:
And the current git state is hosed again

GlazerMann:
@ms609 I have figured out how to do that.

GlazerMann:
If this fails, then we need to yank from Travis and have you trigger it by hand so a human who can fix it is watching

maintain_files.php
```diff
@@ -18,6 +18,7 @@ function git_echo($cmd) {
   git_echo('git config --global user.email "martins@gmail.com"');
   git_echo('git config --global user.name "Martin Smith"');
   git_echo('git fetch --all');
```
ms609:
```suggestion
```

`git pull` contains a `fetch` command; do we need both?

maintain_files.php
```diff
@@ -52,8 +53,11 @@ function git_echo($cmd) {
   git_echo('git add --all *');
   git_echo('git commit -m"Automated file maintenance" || true');
   git_echo('git push https://ms609-bot:' . getenv('GITHUB_PAT') . '@github.com/ms609/citation-bot.git');
+  git_echo('git fetch --all');
+  git_echo('git checkout constants/capitalization.php');
```
ms609: This line looks like it will overwrite the changes that have just been made.  Could the issue be that the `push` is failing to update the GitHub server?

PR #957, opened 2018-10-19
Better support for non-english dates

GlazerMann:
By “support” mean fire napalm at

This is a heavy hammer  for the non-English Whack a mole.

Tests added too (without being asked)

Use "-", not "/", since strtodate assumes Non-american for "-" and american for "/"

Also a few more random things

### DIFF
--- a/expandFns.php
+++ b/expandFns.php
@@ -334,9 +334,7 @@ function tidy_date($string) {
     $string = $cleaned;
   }
   $string = preg_replace("~[^\x01-\x7F]~","-", $string); // Convert any non-ASCII Characters to dashes
-  $string = preg_replace('~\-[\s\-]+~', '-',$string); // Combine dash followed by white or dashes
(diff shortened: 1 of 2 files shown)


expandFns.php
```diff
@@ -320,6 +320,25 @@ function tidy_date($string) {
...
+          $cleaned .= $char;
+       } else {
+          $cleaned .= '-';
+       }
+    }
+    $string = $cleaned;
+  }
+  $string = preg_replace("~[^\x01-\x7F]~","-", $string); // Convert any non-ASCII Characters to dashes
+  $string = preg_replace('~[\s\-]*\-[\s\-]*~', '-',$string); // Combine dash with any following or preceeding white space and other dash
+  $string = preg_replace('~\-$~', '',$string);  // Remove trailing dash
```
ms609: Could we catch multiple dashes (and combine two pregs into one) with:
```suggestion
  $string = preg_replace('~^\-*(.+?)\-*$~', '\1', $string);  // Remove trailing dash
```

PR #959, opened 2018-10-20
Clean google urls

GlazerMann:
https://en.wikipedia.org/wiki/User_talk:Citation_bot#clean_up_google_search_so-called_references

This should probably use explode etc like google books instead


josve05a:
Note, these were just parameters I found in the first 3 Google Search links I found, I bet there are tons more. Maybe they are same/similar parameters as in Google Book links.

GlazerMann:
Need to rewrite

PR #962, opened 2018-10-21
More handles

GlazerMann:


GlazerMann:
@ms609 gotta love GitHub disk failures

GlazerMann:
I think that we just add them as people complain.

PR #968, opened 2018-10-22
Weaken test that fails all the time

GlazerMann:
Accept no date as valid option


GlazerMann:
Basically the test now accepts either good data or nothing.   So, it’s a weaker test

ms609:
Nice solution.

PR #972, opened 2018-10-22
Verify new dois before adding

GlazerMann:


GlazerMann:
Some dois that we get do not resolved.  I am not sure if this is worth doing either.  Perhaps we should count on users using comments to block them.

PR #974, opened 2018-10-23
Verify hdl before tossing url away

GlazerMann:
unregistered handles
Not sure how important now that I removed rogue handle website


Template.php
```diff
@@ -1053,6 +1053,11 @@ protected function get_identifiers_from_url($url_sent = NULL) {
           }
         }
       } elseif (preg_match(REGEXP_HANDLES, $url, $match)) {
+          $url_test = "https://hdl.handle.net/" . urlencode($match[1]);
+          $headers_test = @get_headers($url_test, 1);  // verify that data is registered
+          if ($headers_test !== FALSE && empty($headers_test['Location'])) {
+               return FALSE; // does not resolve.  On FALSE we are optimistic
```
ms609: I don't understand what `On FALSE we are optimistic` means.

GlazerMann: Expanded comment

PR #978, opened 2018-10-24
Support <ref>[htt.... htt..]</ref> expansion

GlazerMann:
When the url title matches the url, try to expand


tests/phpunit/PageTest.php
```diff
@@ -69,6 +69,8 @@ public function testUrlReferences() {
       $this->assertEquals("URL reference test 1 <ref name='bob'>{{Cite journal | doi=10.1007/s12668-011-0022-5|title = Reoccurring Patterns in Hierarchical Protein Materials and Music: The Power of Analogies| journal=Bionanoscience| volume=1| issue=4| pages=153–161|year = 2011|last1 = Giesa|first1 = Tristan| last2=Spivak| first2=David I.| last3=Buehler| first3=Markus J.| arxiv=1111.5297}}< / ref>\n Second reference: \n<ref >{{Cite journal |pmc = 3705692|year = 2013|last1 = Mahajan|first1 = P. T.|title = Indian religious concepts on sexuality and marriage|journal = Indian Journal of Psychiatry|volume = 55|issue = Suppl 2|pages = S256–S262|last2 = Pimple|first2 = P.|last3 = Palsetia|first3 = D.|last4 = Dave|first4 = N.|last5 = De Sousa|first5 = A.|pmid = 23858264|doi = 10.4103/0019-5545.105547}}</ref> URL reference test 1", $page->parsed_text());
       $page = $this->process_page(" text <ref name='dog' > 10.1063/1.2263373 </ref>");
       $this->assertTrue((boolean) strpos($page->parsed_text(), 'title'));
+      $page = $this->process_page(" text <ref name='dog' >[http://10.1007/s12668-011-0022-5 http://10.1007/s12668-011-0022-5]</ref>");
```
ms609: This test implies that the new functionality is solely to handle malformed input (http://doi).  It would be better to use a valid URL as an example, if that is the intention.

GlazerMann: That was stupid copy/paste error

PR #981, opened 2018-10-25
Replace title={title} with real data

GlazerMann:
more common than it should be.......


GlazerMann:
The next one I add I will make an array

PR #982, opened 2018-10-25
More google domains

GlazerMann:
google.com.au


GlazerMann:
You are right.  This bug report was a misunderstanding.

PR #983, opened 2018-10-25
Remove work=google books, like we do publisher

GlazerMann:


GlazerMann:
I will work on array.  I might make the bad title = pretend blank too

PR #995, opened 2018-10-28
Remove via= with no url for pubmed & JSTOR

GlazerMann:
The bot used to not remove via=, thus it looks like a lot of PubMed and PubMed Central Vias need removed.

Refill adds via= when expanding URLs.  We then converted them to IDs and now we need to get rid of all those vias


ms609:
Can we add a test, to satisfy codecov?

PR #1005, opened 2018-10-31
Only this file

GlazerMann:
Less likely to goon git repository if we only commit the one file we want.

Also, then update that file — should be our changes — to avoid “uncommitted changes” error.


ms609:
I still believe that the problem is that the changes to this file are not being committed to the GitHub repository by ms609-bot.  I've still not had time to look into why this is the case.

PR #1007, opened 2018-10-31
Do not grab last edited dates some sites

GlazerMann:
Added to profoundly_incomplete() so that we do not mind lack of date.
Do not get saved date from archives.
Do not get last edited date from wikipedia

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -303,7 +303,7 @@ final class Template {
     // Zotero translation server often returns bad data, which is worth having if we have no data,
     // but we don't want to fill a single missing field with garbage if a reference is otherwise well formed.
     $has_date = $this->has("date") || $this->has("year") ;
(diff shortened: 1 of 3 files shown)


Template.php
```diff
@@ -299,23 +299,31 @@ public function incomplete() {
     ));
   }
 
-  public function profoundly_incomplete() {
+  public function profoundly_incomplete($url = '') {
     // Zotero translation server often returns bad data, which is worth having if we have no data,
     // but we don't want to fill a single missing field with garbage if a reference is otherwise well formed.
+    $has_date = $this->has("date") || $this->has("year") ;
+    foreach (NO_DATE_WEBSITES as $bad_website ) {
```
ms609:
```suggestion
    foreach (NO_DATE_WEBSITES as $bad_website) {
```

PR #1012, opened 2018-11-01
fix: fix open-access url when oadoi returns no url_for_landing_page

monperrus:
fix open-access url when oadoi returns no url_for_landing_page

- check for null value in JSON field url_for_landing_page
- prefer PDF url is one exists


GlazerMann:
We do not prefer PDF, because they are historically slightly less reliable

monperrus:
thanks a lot! --Martin

PR #1029, opened 2018-11-05
zenodo.org has been blacklisted

josve05a:
https://en.wikipedia.org/w/index.php?title=User_talk:Citation_bot&oldid=867463354#Emergency_blacklist


nemobis:
Note that it's supposed to be temporary

PR #1036, opened 2018-11-06
Prepare to  Fix JSTOR

GlazerMann:
updates code to use curl and makes it ready to go

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -515,11 +515,11 @@ function expand_by_jstor($template) {
   if (stripos($dat, 'No RIS data found for') !== FALSE) {
     report_info("JSTOR API found nothing for ".  jstor_link($jstor));
     return FALSE;
-  }  
+  }
   if (stripos($dat, 'Block Reference') !== FALSE) {
     report_info("JSTOR API blocked bot for ".  jstor_link($jstor));
     return FALSE;
-  }   
+  }
   $has_a_url = $template->has('url');
   $template->expand_by_RIS($dat);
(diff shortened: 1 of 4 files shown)


ms609:
Well, we are a robot!
Do we need to get an API key?  How might we obtain one?

GlazerMann:
I got a UUID from my FireFox.  I highly doubt that it will continue to work, but it points to the possibility of working.

GlazerMann:
I will e-mail JSTOR

apiFunctions.php
```diff
@@ -492,7 +492,26 @@ function expand_by_jstor($template) {
...
+      'header'=>"Accept-language: en\r\n" .
+                "Cookie: UUID=3d1209ba-d7d6-40dc-8f1a-9190c982e0f2\r\n"
+    )
+    );
+  } else {
+   $jstor_opts = array(
+    'http'=>array(
+      'method'=>"GET",
+      'header'=>"Accept-language: en\r\n" .
+                "Cookie: UUID=2c114fa7-67a8-4291-a967-e4c5c5a563ac\r\n"
```
ms609: This cookie will only exist on your local machine.  CURL has cookie-handling facilities.  But the appropriate thing here is to use JSTOR's API; if this route is not designed for programmatic access, then we should respect JSTOR's wishes and not overload their servers.

GlazerMann: I have an email into JSTOR.  @ms609 we can commit as-is for now which JSTOR on the toolforge turned off.

GlazerMann:
CURL does not seem to work...

PR #1037, opened 2018-11-06
do not remove archive-url, if another url is present

GlazerMann:


GlazerMann:
Our code already policed them by converting direct links to files.

GlazerMann:
Sorry for changing this pull, but Zenodo is not coming back anytime soon.  And, we are being destructive.

PR #1047, opened 2018-11-08
Prepare to Do researchgate ourself

GlazerMann:
Right now just sends FALSE back, which is better since we don’t waste time.  Code is ready for key

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -509,6 +509,35 @@ function expand_by_jstor($template) {
   return TRUE;
 }
 
+function query_researchgate_api($ids, $templates) {
+  foreach ($templates as $template) expand_by_researchgate($template);
+}
+
+function expand_by_researchgate($template, $url_sent = NULL) {
+  if ($template->incomplete() === FALSE) return FALSE;
+  if (is_null($url_sent)) $url_sent = $template->get('url');
+  if (preg_match("~^https?://(?:www.|)researchgate.net/[^\s]*publication/([0-9]+)~i", $url_sent, $match)) {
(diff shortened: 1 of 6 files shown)


apiFunctions.php
```diff
@@ -509,6 +509,38 @@ function expand_by_jstor($template) {
...
+  } else {
+    return FALSE;
+  }
+  $cookie_opts = array(
+    'http'=>array(
+      'method'=>"GET",
+      'follow_location' => true,
+      'header'=>"Accept-language: en\r\n" .
+                "Cookie: cookieconsent_dismissed=true\r\n" . 
+                "User-Agent: Mozilla/5.0 (Citation Bot)\r\n" .
```
ms609: Can we use a globally defined User-Agent?  We ought to have this in a constant somewhere...

apiFunctions.php
```diff
@@ -509,6 +509,38 @@ function expand_by_jstor($template) {
...
+    $url = 'https://www.researchgate.net/publicliterature.PublicationHeaderDownloadCitation.downloadCitation.html?publicationUid=' . $match[1] . '&fileType=RIS&citationAndAbstract=false'; // Convert researchgate URL to give RIS information
+  } else {
+    return FALSE;
+  }
+  $cookie_opts = array(
+    'http'=>array(
+      'method'=>"GET",
+      'follow_location' => true,
+      'header'=>"Accept-language: en\r\n" .
+                "Cookie: cookieconsent_dismissed=true\r\n" .
```
ms609: Can we handle cookies more elegantly using CURL?

PR #1048, opened 2018-11-08
 Do not add OA url if free ID exists

josve05a:
"Some identifiers always link to free full texts. In this case, the access level is automatically indicated by the template. This is the case for |arxiv=, |biorxiv=, |citeseerx=, |pmc=, |rfc= and |ssrn=."

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1768,7 +1768,7 @@ final class Template {
           }
         }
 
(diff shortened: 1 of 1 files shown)


josve05a:
Please add the requested things from https://en.wikipedia.org/wiki/User_talk:Citation_bot#Request:_Do_not_add_OA_url_if_|foo-access=free as well

GlazerMann:
Those are often pre prints etc

josve05a:
> Those are often pre prints etc

Our OA urls are ofter submitted manuscripts and pre-prints as well

PR #1056, opened 2018-11-13
Add data sanity checks

GlazerMann:
Paranoid.  Just do not want us to try to deal with bad data.  Help prevent deliberate/accidental DOS attacks.  With the current code, calling with no arguments is a DOS -- See infinite loop bug.

### DIFF
--- a/generate_template.php
+++ b/generate_template.php
@@ -8,16 +8,16 @@ header("Content-Type: text/plain");
 include('expandFns.php');
 $t = new Template();
 $t->parse_text('{{cite web}}');
-if (count($_GET) > 10) die('Excessive parameters passed');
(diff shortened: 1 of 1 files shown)


generate_template.php
```diff
@@ -8,7 +8,14 @@
 include('expandFns.php');
 $t = new Template();
 $t->parse_text('{{cite web}}');
-foreach ($_GET as $param=>$value) $t->set($param, $value);
+$count = 0;
```
ms609:
```suggestion
if (count($_GET) > 10) fail_safely() //?
```

ms609: Can `die` die with some form of error message to the user?
It's not quite clear what's being screened for here.

PR #1062, opened 2018-11-14
Simplify Jstor tests

GlazerMann:
Only care about loss of url and only having one jstor= block.
Thus, merge into a different test.

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -40,7 +40,7 @@ final class TemplateTest extends testBaseClass {
     $prepared = $this->prepare_citation($text);
     $this->assertEquals('40237667', $prepared->get('jstor'));
     $this->assertNull($prepared->get('doi'));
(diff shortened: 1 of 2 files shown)


tests/phpunit/TemplateTest.php
```diff
@@ -1122,7 +1122,7 @@ public function testJstorSICIEncoded() {
   public function testOverwriteBlanks() {
     $text = '{{cite journal|url=http://www.jstor.org/stable/1234560|jstor=}}';
     $expanded = $this->process_citation($text);
-    $this->assertEquals('{{cite journal|jstor=1234560}}', $expanded->parsed_text());
+    $this->assertEquals(1, substr_count($expanded->parsed_text(), 'jstor'));
```
ms609: But won't this return success if the text is not expanded, by virtue of the `jstor` in the URL?

GlazerMann: This test is not to see if it expanded or not, but if we remember to not leave the blank jstor= too.  The code at one point did that

PR #1066, opened 2018-11-14
Better String Comparison

GlazerMann:
We do this in three places so create a function.  Also add ignoring leading "the"

### DIFF
--- a/expandFns.php
+++ b/expandFns.php
@@ -429,9 +429,5 @@ function str_almost_the_same($str1, $str2) {  // Move forgiving than ===
   $str2 = trim($str2);
   $str1 = preg_replace("~^the\s+~", "", $str1);  // Ignore leading "the" so "New York Times" == "The New York Times"
   $str2 = preg_replace("~^the\s+~", "", $str2);
-  if ($str1 === $str2) {
-    return TRUE;
-  } else {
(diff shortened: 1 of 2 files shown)


expandFns.php
```diff
@@ -419,3 +419,15 @@ function equivalent_parameters($par) {
     default: return array($par);
   }
 }
+
+function str_almost_the_same($str1, $str2) {  // For comparing strings with forgiveness
+  $str1 = mb_strtolower($str1); // Case-insensitive
```
ms609: If you are doing the same thing to $str1 and $str2, could you create a separate function to treat them?  This avoids unnecessary code repetition. 
Then `str_almost_the_same` could be simply `return(strcasecmp(remove_irrelevant_bits($str1), remove_irrelevant_bits($str2)) == 0)`

expandFns.php
```diff
@@ -419,3 +419,15 @@ function equivalent_parameters($par) {
     default: return array($par);
   }
 }
+
+function str_almost_the_same($str1, $str2) {  // For comparing strings with forgiveness
```
ms609: This is a slightly opaque name for the function; it's not quite clear how "almost" is "almost".  Would `str_equivalent` be clearer?

PR #1072, opened 2018-11-15
Use dx.doi.org specific code before generic DOI code

GlazerMann:
move dx.doi.org check before generic doi checker -- fixed this bug
add test for two dois that does not currently drop properly  -- this bug
changed white space output for page test -- unintended side effect of code path change


GlazerMann:
preg_match("~^https?://(?:d?x?\.?doi\.org|doi\.library\.ubc\.ca)/([^\?]*)~i", $url, $match)) {
        quietly('rep

PR #1079, opened 2018-11-16
do not expand google searches

GlazerMann:
Also make jstor regex non-capturing, since we don't capture it


ms609:
Is it time to make an array of matches to test against?

GlazerMann:
Next time I make an array.  Maybe later today

PR #1080, opened 2018-11-16
simplify google search URLS

GlazerMann:
Just like we do to google books, researchgate, etc.


GlazerMann:
--- Expected
+++ Actual
@@ @@
-'https://www.google.com/search?q=%22institute+for+sustainable+weight+loss%22&oq=%22institute+for+sustainable+weight+loss%22'
+'https://www.google.com/search?https://www.google.com/search&q=%22institute+for+sustainable+weight+loss%22'

PR #1113, opened 2018-12-12
remove "Submitted manuscript"

monperrus:
For many accepted papers, the note "Submitted manuscript" is confusing. Eg
https://tools.wmflabs.org/citations/generate_template.php?doi=10.1145/3238147.3240474Fixes

- discard unreliable info "Submitted manuscript"


monperrus:
I don't think so, but I don't know all possible usages of this field in the wild.

GlazerMann:
Cue the hordes of editors with torches and pitchforks

GlazerMann:
This pull assumes that all format data is invalid.  Even format=Excel, PDF, Liner notes, etc

PR #1122, opened 2018-12-16
Match an URL as redundant with DOI even with extra fragments

nemobis:
The documentation for extract_doi() states that the output may include
some leftovers which don't belong to the DOI proper, so a perfect match
may fail.

Example: https://en.wikipedia.org/?diff=873887190


GlazerMann:
The test already works in the existing code
Also, you want to change the strcasecmp line instead I think

nemobis:
Thanks for checking.

GlazerMann, 16/12/18 21:30:
> The test already works in the existing code

Yet it doesn't work in the live tool.

> Also, you want to change the strcasecmp line instead I think

Right, but I'm not sure what that line is doing. Does it only want to 
check that the DOI string extracted from the URL is identical to the DOI?

nemobis:
Martin R. Smith, 20/12/18 11:45:
> *@ms609* approved this pull request.

Thanks. The next patch I'd like to do, if there's interest, is to follow 
redirects from publishers which have a DOI resolver: there are a few 
hundreds.
https://www.crossref.org/blog/urls-and-dois-a-complicated-relationship/

It's necessary to accept cookies but we can do without JavaScript and 
identify the destination with some custom logic, at least for the 
biggest players (mostly Elsevier).

PR #1145, opened 2018-12-28
Better support for non-journal websites

GlazerMann:
We hit websites like CNN.com over and over again even when we do not have to and will gain nothing.


GlazerMann:
Trailing dash reduces false positives

PR #1147, opened 2018-12-28
(New Series) not (new Series)

GlazerMann:


GlazerMann:
Too many counter examples in titles come to mind

PR #1156, opened 2019-01-01
Doi hidden in “extra”

GlazerMann:
Some translators just dump stuff in "extra". 
Error checking and data verification done.
@ms609 changes made to guarantee only do this if one DOI.


GlazerMann:
Need to:
verify that we only get one match
Add code to fix error from other test.
Think about

ms609:
I worry that this is going to take a fair bit of maintenance and that the "extra" field may not have a constant format, but if you're willing to keep on top of new occurrences as they arise, then why not make use of the data!

PR #1171, opened 2019-01-06
Do not add banned urls

GlazerMann:
we can’t save if we add these
these are from non-cross-ref doi's

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -735,6 +735,7 @@ final class Template {
       ### arXiv, DOI, PMID etc. ###
       
       case 'url': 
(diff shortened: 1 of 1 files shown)


josve05a:
Perhaps to future-saving ourself, we should perhaps make a copy of (or fetch) the official blacklist of links. In case more links are to be banned, or for when this is removed from the blacklist.

GlazerMann:
There are actually multiple ever changing blacklists.  Do you know a good way to find them?

josve05a:
https://en.wikipedia.org/wiki/MediaWiki:Spam-blacklist is the main one

GlazerMann:
Fixed elsewhere in code

PR #1174, opened 2019-01-06
Verify that title matches in bibcode search

GlazerMann:
We sometimes get things that are not right

We should not add bibcodes if the title does not match

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1482,6 +1482,21 @@ final class Template {
           return FALSE;  // Probably a book review or something with same title, etc.
         }
       }
+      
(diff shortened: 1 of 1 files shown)


Template.php
```diff
@@ -1482,6 +1482,21 @@ public function expand_by_adsabs() {
           return FALSE;  // Probably a book review or something with same title, etc.
         }
       }
+      
+      if ($this->has('title')) { // Verify the title matches.  We get so strange mis-matches
```
ms609:
```suggestion
      if ($this->has('title')) { // Verify the title matches.  We get some strange mis-matches
```

PR #1178, opened 2019-01-06
detect corrupt big queries

GlazerMann:
http://adsabs.harvard.edu/abs/1982MSS...C03....0H changes to another bibcode and the whole query results are wrong


GlazerMann:
@ms609 perhaps since you have the key, you could debug and provide feeeback to adsabs

PR #1212, opened 2019-01-11
Do not mess with dashes in [// URLs and comments

GlazerMann:
Forgot this one

add a comment in both locations pointing to the other location 

also add  CITATION_BOT_PLACEHOLDER_COMMENT to Template.php

The orginal tests I committed were not valid :-(


ms609:
Is `can_safely_change` sufficiently specific -- would `can_safely_modify_dashes` be more precise?

PR #1223, opened 2019-01-13
Avoid returning "0" as page number/range

nemobis:
Example:
https://en.wikipedia.org/?diff=878079480

Coming from `"page":"0-0"` in:
https://api.crossref.org/works/10.1111/j.1755-3768.2010.4142.x

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -184,7 +184,7 @@ final class Template {
   }
   
   public function process() { // This code used to be the core of the Citation Bot, now it is only used by generate_template.php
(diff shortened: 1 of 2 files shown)


GlazerMann:
Modified it too work.  Also, added code that marks a test as not runnable by non-team-members (that's a bug I introduced).

GlazerMann:
Deleted a duplicate test that I made at same time.

ms609:
I have a vague recollection of doing a similar test (for `0`, but not `0-0`) when validating CrossRef data; it may be worthwhile reviewing this test too, to avoid an unnecessary call to "add_if_new".

PR #1225, opened 2019-01-13
Better doi validation

GlazerMann:
The current code changes the correct DOI
10.1126/science.10.1126/SCIENCE.291.5501.24
to something else because we see 10.1126/science.10.1126/SCIENCE.291.5501.24 as wrong.
It is rubbish, but it is valid.  What were they thinking?

The current code has issues if a DOI contains "10.####/ somewhere in it.

@ms609 thoughts?


GlazerMann:
@ms609 thoughts on this?
Should it go on end instead of on front?

PR #1226, opened 2019-01-16
Some small non-controversial fixes

nemobis:
Spanish doesn't use Title Case.
https://en.wikipedia.org/?diff=878682078


nemobis:
Is it mandatory to have every name in both arrays? I feel it's fine to leave alone the citations where the name has already been capitalised.

GlazerMann:
Yes, it needs to be in both arrays because it is a change this to that system.

Secondly style guides generally say to capitalize Spanish journals when writing in English

GlazerMann:
Even a lot of Spanish style guides say to capitalize journal titles—they are called out as an exception in Spanish

PR #1236, opened 2019-01-18
Catch more bare DOI citations and send them to cleanup

nemobis:


ms609:
Thanks for the pull request; it'd be really handy if you could add some tests to the test suite that will invoke the code you've written; i.e. dois that fail under the existing code that will be handled correctly under your proposal.

nemobis:
I agree, I was about to post some examples: https://phabricator.wikimedia.org/P8007
Do you have suggestions on which ones to include?

GlazerMann:
We should have at least one for each Regex

GlazerMann:
<ref>John Doe, [https://doi.org/10.1244/abc Foo], Bar 1789.</ref> regex is messed up.  I fixed some issues, but not all

GlazerMann:
I cancelled the 5.6 build since we have more code to write

GlazerMann:
I added a bunch of tests.  They all will fail, since I do not know the output yet.

PR #1238, opened 2019-01-19
Dashes are okay issues too

GlazerMann:
issue 2-3 becomes -1 when cast to integer

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -399,7 +399,7 @@ function expand_by_doi($template, $force = FALSE) {
       $template->add_if_new('isbn', $crossRef->isbn);
       $template->add_if_new('journal', $crossRef->journal_title); // add_if_new will format the title
       if ($crossRef->volume > 0) $template->add_if_new('volume', $crossRef->volume);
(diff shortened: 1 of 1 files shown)


apiFunctions.php
```diff
@@ -399,7 +399,7 @@ function expand_by_doi($template, $force = FALSE) {
       $template->add_if_new('isbn', $crossRef->isbn);
       $template->add_if_new('journal', $crossRef->journal_title); // add_if_new will format the title
       if ($crossRef->volume > 0) $template->add_if_new('volume', $crossRef->volume);
-      if ((integer) $crossRef->issue > 1) {
+      if (((integer) $crossRef->issue > 1) || (strpos($crossRef->issue, '-') > 0)) {
```
ms609: Perhaps better practice do to in the opposite sequence:
```suggestion
      if (((strpos($crossRef->issue, '-') > 0 || (integer) $crossRef->issue > 1))) {
```
so that a hyphen will never be cast to an integer (by lazy evaluation).

PR #1250, opened 2019-01-22
We need this to not be the default

GlazerMann:
we some way to make this be a checkbox in the GUI

Beware of pitchforks 

I think when submitting a single page we need another option like “slow” to turn this on.   Perhaps only in gadget API so people always know who did it.

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -141,7 +141,7 @@ class Page {
                       substr_count($this->text, '{{Cite ') +
                       substr_count($this->text, '{{citation') +
                       substr_count($this->text, '{{Citation');
-    $ref_count = substr_count($this-text, '<ref') + substr_count($this-text, '<Ref');
+    $ref_count = substr_count($this->text, '<ref') + substr_count($this->text, '<Ref');
     // PLAIN URLS Converted to Templates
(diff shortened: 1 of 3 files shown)


Page.php
```diff
@@ -167,6 +172,10 @@ function($matches) {return $matches[1] . '{{Cite journal | id=' . $matches[2] .
                       function($matches) {return $matches[1] . '{{Cite journal | doi=' . $matches[2] . ' | ' . strtolower('CITATION_BOT_PLACEHOLDER_BARE_URL') .'=' . base64_encode($matches[2]) . '}}' . $matches[3] ;},
                       $this->text
                       );
+     if (
+        ($ref_count < 2) ||
+        (($citation_count/$ref_count) >= 0.5)
+     ) {
      $this->text = preg_replace_callback(   // like <ref>John Doe, [https://doi.org/10.1244/abc Foo], Bar 1789.</ref>
                                             // also without titles on the urls
```
nemobis: The case without title, where there's only a link in the ref, should not be within this condition.

GlazerMann:
Do not comment on closed pull requests 
No ones reads this.
You got lucky this time

PR #1260, opened 2019-01-26
Remove URLs that match DOI

GlazerMann:
@ms609 any thoughts?

First, the code makes an informed guess about whether this is even reasonable to try (DOI work, citation is complete, URL is a know publisher website, etc).

The code finds where dx.doi.org ends up pointing to.
The code finds where the $url ends up pointing to (because it might have moved, or as is often the case, the link is to a pay-money page, and the redirect is to the free-preview page).
The code then also removes anything after a ? or # in the URLs.

Checks to see if the DOI based URL matches either the $url or the redirectred $url

### DIFF
--- a/api_handlers/zotero.php
+++ b/api_handlers/zotero.php
@@ -24,6 +24,20 @@ function query_url_api($ids, $templates) {
        if ($template->has('rfc'))       expand_by_zotero($template, 'https://tools.ietf.org/html/rfc' . $template->get('rfc'));
        if ($template->has('ssrn'))      expand_by_zotero($template, 'https://papers.ssrn.com/sol3/papers.cfm?abstract_id=' . $template->get('ssrn'));
        if ($template->has('doi') && !doi_active($template->get('doi')))  expand_by_zotero($template, 'https://dx.doi.org/' . urlencode($template->get('doi'))); // Non-crossref DOIs, such as 10.13140/RG.2.1.1002.9609
+    // This is in wrong place
+    if ($template->has('doi') &&
+        !$template->incomplete() &&
+        $template->get('url') === $url &&
+        !preg_match(REGEXP_DOI_ISSN_ONLY, $template->get('doi')) &&
+        str_ireplace(CANONICAL_PUBLISHER_URLS, '', $url) != $url && // This is the use a replace to see if a substring is present trick
+        str_ireplace(['pdf', 'image', 'plate', 'figure', 'picture'], '', $url) == $url && // might be a link to a specific image or direct pdf link
+        $template->blank(DOI_BROKEN_ALIASES) &&
+        doi_active($template->get('doi'))) // check one that accesses network last
+      // DOI Active needs changed to be a dx.doi.org check
+    {
+          report_forget("Existing canonical URL resulting from equivalent DOI; dropping URL");
+          $template->forget('url');
+    }
   }
 }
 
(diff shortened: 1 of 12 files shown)


api_handlers/zotero.php
```diff
@@ -25,6 +25,49 @@ function query_url_api($ids, $templates) {
        if ($template->has('ssrn'))      expand_by_zotero($template, 'https://papers.ssrn.com/sol3/papers.cfm?abstract_id=' . $template->get('ssrn'));
        if ($template->has('doi') && !doi_active($template->get('doi')))  expand_by_zotero($template, 'https://dx.doi.org/' . urlencode($template->get('doi'))); // Non-crossref DOIs, such as 10.13140/RG.2.1.1002.9609
   }
+   
+  // Now that we have expanded URLs, try to lose them
+  $ch = curl_init();
+  curl_setopt($ch, CURLOPT_FOLLOWLOCATION, TRUE);
+  curl_setopt($ch, CURLOPT_MAXREDIRS, 100); // No infinite loops for us
```
nemobis: 10 or even 5 is plenty,

api_handlers/zotero.php
```diff
@@ -25,6 +25,49 @@ function query_url_api($ids, $templates) {
...
+            $url_short         = strtok($url,               '?#');
+            if (0 === strcasecmp($url_short, $redirectedUrl_doi)) {
+               report_forget("Existing canonical URL resulting from equivalent DOI; dropping URL");
+               $template->forget('url');
+            } else { // See if $url redirects
+               curl_setopt($ch, CURLOPT_URL, $url);
+               if (@curl_exec($ch)) {
+                  $redirectedUrl_url = curl_getinfo($ch, CURLINFO_EFFECTIVE_URL);
+                  $redirectedUrl_url = strtok($redirectedUrl_url, '?#');
+                  if (0 === strcasecmp($redirectedUrl_url, $redirectedUrl_doi)) {
```
nemobis: Better check that the resolved URL is contained in the linked URL, rather than identical, to avoid false negatives due to suffixes, anchors and redundant query parameters.

api_handlers/zotero.php
```diff
@@ -25,6 +25,49 @@ function query_url_api($ids, $templates) {
...
+    $url = $template->get('url');
+    if ($doi &&
+        $url &&
+        !$template->incomplete() &&
+        !preg_match(REGEXP_DOI_ISSN_ONLY, $doi) &&
+        str_ireplace(CANONICAL_PUBLISHER_URLS, '', $url) != $url &&
+        str_ireplace(['pdf', 'image', 'plate', 'figure', 'picture'], '', $url) == $url &&
+        $template->blank(DOI_BROKEN_ALIASES))
+    {
+          curl_setopt($ch, CURLOPT_URL, "https://dx.doi.org/" . urlencode($doi));
```
nemobis: Why do you even make an actual GET with curl, rather than a simple HEAD request? If your objective is to follow redirects made by the publishers' DOI resolvers, this is impossible for many of them unless you also enable cookies. This may not be the default on all versions of curl.
https://curl.haxx.se/docs/http-cookies.html

A test may be useful to check whether you're getting the dreaded cookieAbsent page, e.g.:
https://www.journals.uchicago.edu/action/cookieAbsent
https://www.tandfonline.com/action/cookieAbsent
https://www.annualreviews.org/action/cookieAbsent
https://aip.scitation.org/action/cookieAbsent

api_handlers/zotero.php
```diff
@@ -25,6 +25,49 @@ function query_url_api($ids, $templates) {
...
+          if (@curl_exec($ch)) {
+            $redirectedUrl_doi = curl_getinfo($ch, CURLINFO_EFFECTIVE_URL);  // Final URL
+            $redirectedUrl_doi = strtok($redirectedUrl_doi, '?#');  // Remove session stuff
+            $url_short         = strtok($url,               '?#');
```
nemobis:
```suggestion
            $url_short         = strtok($url,               '?#');
            if ( preg_match('~https://linkinghub.elsevier.com/retrieve/pii/(S[0-9]+)~i', $redirectedUrl_doi, $matches ) === 1 ) {
                 $redirectedUrl_doi = 'https://www.sciencedirect.com/science/article/pii/' . $matches[1] ;
             }
```

nemobis: 20 % of DOIs warrant a special matching for Elsevier. This is a simplistic solution in case you want to only use HTTP headers. Otherwise you'd need to match the actual content of the final HTML, looking for a string like `Redirect=https%3A%2F%2Fwww.sciencedirect.com%2Fscience%2Farticle%2Fpii%2FS0167923612002370`, à la

```
preg_match('~Redirect=(https[^>]+pii%2FS0167923612002370)~', $html, $matches);
$redirectedUrl_doi = urldecode($matches[1]);
```

PR #1281, opened 2019-01-31
Do not use expand_by_doi() twice

GlazerMann:
@ms609  Fixed in within the template.  Call it all you want, we will only expand it once, unless the doi changes.  So now we can safely call it three times (first, last, and in the middle).  Also, the arXiv call when it finds a DOI is covered too.

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -214,6 +214,10 @@ class Page {
       }
     }
     
+    $our_dois = array();
+    for ($i = 0; $i < count($our_templates); $i++) {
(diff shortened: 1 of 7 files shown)


GlazerMann:
@ms609  You are right, I mis-read the code.  It is now fixed

GlazerMann:
There are still a few code paths where is will be called twice, but much fewer of them (The arXiv code for example immediatly does it since the DOI is considered better than the arXiv data).

PR #1297, opened 2019-02-07
Do not require https or /article for match

nemobis:
The match against the existing URL fails if it's http or a chapter.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -385,12 +385,13 @@ final class Template {
         if ($this->blank(DISPLAY_AUTHORS)) {
           return $this->add($param_name, $value);
         }
-      return FALSE;
+        return FALSE;
+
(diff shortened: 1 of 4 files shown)


nemobis:
Hmm, here a http URL was removed anyway https://en.wikipedia.org/w/index.php?title=Latrodectus&diff=prev&oldid=882236844

nemobis:
Maybe we should just match the Elsevier ID, so that the odd URLs like http://www.thelancet.com/pdfs/journals/lancet/PIIS014067360211244X.pdf would also be covered.

GlazerMann:
code changed to convert both common kinds of PII urls.
code chnaged to convert all https to http, just so that matches are more likely

PR #1302, opened 2019-02-07
convert last3=et. al. to display-authors = etal

GlazerMann:


GlazerMann:
@ms609 the display=etal is fairly new.  This only fixes cases where editors list the last author at etal.

PR #1317, opened 2019-02-09
convert tabs, linefeeds, null bytes to spaces

GlazerMann:


GlazerMann:
@ms609 blocked it from effecting quote= and urls

PR #1320, opened 2019-02-09
Stop hammering Zenodo for generic URLs when it keeps failing

nemobis:
Avoids this kind of tedious and pointless case where we fail a hundred times in a row:
https://phabricator.wikimedia.org/P8064

### DIFF
--- a/api_handlers/zotero.php
+++ b/api_handlers/zotero.php
@@ -1,16 +1,12 @@
 <?php 
+global $zotero_failures_count;
+
 function query_url_api($ids, $templates) {
+  $zotero_failures_count = 0;
   report_action("Using Zotero translation server to retrieve details from URLs.");
-  $failures = 0;
   foreach ($templates as $template) {
     if ($template->has('url')) {
       $response = expand_by_zotero($template);
-    }
-    if ($response === FALSE) $failures += 1;
-    if ($failures >= 5 ) {
-      // Repeat failures for errors on our side: probably a chronic timeout.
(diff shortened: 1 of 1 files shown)


api_handlers/zotero.php
```diff
@@ -10,7 +10,7 @@ function query_url_api($ids, $templates) {
     if ($failures >= 5 ) {
       // Repeat failures for errors on our side: probably a chronic timeout.
       report_action("Giving up on generic URLs for this page.")
-      break;
+      return;
```
nemobis: No, this is not what I intended. In my example, the rest of the operation continued correctly. Please don't skip the entire thing.

api_handlers/zotero.php
```diff
@@ -144,7 +142,7 @@ function expand_by_zotero(&$template, $url = NULL) {
       return FALSE;
     case 'Remote page not found':
       report_info("Remote page not found for URL ". $url);
-      return NULL;
+      return FALSE;
   }
   
   if (strpos($zotero_response, '502 Bad Gateway') !== FALSE) {
```
nemobis: My idea was to count 502 errors from Tools and the other failures on our side, but ok.

GlazerMann:
The only thing that really points to the tool having trouble is the timeouts

GlazerMann:
@nemobis how about this.  We keep the dropping of urls code and such.

nemobis:
Sure, this could be enough. I'm not sure if the global is going to cause trouble later.

PR #1322, opened 2019-02-09
improve URL=DOI code

nemobis:
* Do not require URL to match some narrow regexes before resolving redirects.
* Fix regex typo introduced in 36c462860a6b003fd424bc3b6a7b98b1c962f740.


nemobis:
Examples from https://en.wikipedia.org/w/index.php?title=Phlegraean_Fields&diff=prev&oldid=882518580

GlazerMann:
The reason for the narrowness initialialy was to make sure that if it did anything bad, it would only do so in the case where there was a doi anyway, so nothing important was lost.

PR #1331, opened 2019-02-12
Detect errors rather than crash

GlazerMann:
actually seen it fail once


nemobis:
Are we not sending a contact address and descriptive user agent?
https://github.com/CrossRef/rest-api-doc#etiquette

GlazerMann:
Yes, we do.  But, I saw this one time out of thousands, so I added this pull

PR #1335, opened 2019-02-13
User OAUTH code

GlazerMann:
@ms609  needs to upgrade the exising OAuth tokens so that they can also get user tokens at https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration.  Once they are valid for users.  Then we get to debug on the development branch.  You will also have to specify a return URL: I chose authenticate.php

Unknown:  Do we need a new edit token for editing of a file.  ie. do we need to grab a new edit token for each and every file in a category?

Once we can do that, we should tag our edits (this is can be setup to automatically do this based upon our oauth keys) https://en.wikipedia.org/wiki/Special:Tags


pppery:
If you're using OAuth, wouldn't it be better to go all-out and actually make the edit using the activating user's account instead of the bot's?

GlazerMann:
That’s exactly what this pull does @pppery

pppery:
Sorry, I'm not very good at reading complicated PHP code.

GlazerMann:
@ms609 people are grumpy about the anonymous nature and I am using it as a learning opportunity.  No rush at all.  Some more ideas hit me recently on this and I have some changes to make.

GlazerMann:
@ms609 there is a movement to PERMANENTLY BLOCK us unless be stop all anonymous editing.  We need to get this working on the dev bot.

GlazerMann:
https://en.wikipedia.org/w/index.php?title=Wikipedia:Bot_policy&diff=next&oldid=204147936 

The bots current behavior violates Wikipedia policy that changed a decade ago

GlazerMann:
TODO: Dump all the header stuff and force people to click a link

ms609:
I'm not sure how to upgrade OAuth tokens; the 'manage tokens' page doesn't include such an option.  Could you provide a link?

GlazerMann:
I think you have to go to https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration as the citation bot account.  Not 100% sure

category.php
```diff
@@ -47,7 +47,7 @@
 <?php
 }
 
-$user = isset($_REQUEST["user"]) ? $_REQUEST["user"] : NULL;
+$user = NULL; // Editing as user or script, no need to name.
 if (is_valid_user($user)) {
   echo " Activated by $user.\n\n";
   $edit_summary_end = " | [[User:$user|$user]]; [[Category:$category]].";
```
pppery: This is dead code, and, once the edits are done using OAUTH, there is no need to add "user-activated" to the edit summary (alternately, replace it with something like "using Citation bot")

GlazerMann: Original attempt was to minimize code changes

PR #1341, opened 2019-02-16
EMERGENCY: Do not slam BIBCODE key on Travis

GlazerMann:
project members have it set in TRAVIS and just hammer on it.   My apologies.

@ms609 I think we need this because I did too many pull edits and the bot is being run **a lot**

Also, make the tests fail when we have used up our welcome


ms609:
I'm not quite clear why `getenv('TRAVIS')` isn't sufficient?  I'd prefer to avoid global variables where possible.

GlazerMann:
I was thinking that you might run the secret tests on Travis @ms609

GlazerMann:
@ms609 Your suggestions implemented

PR #1364, opened 2019-02-22
Red Alert:  Location and publisher gets to stay

GlazerMann:
the RFC went our way mostly, but not quite enough for a bot to do the work


Sky2042:
"You should probably remove" is not really appropriate given the close of the RFC.

GlazerMann:
The close was that it was probably best to remove them; but, by hand since is rare cases they would be information

Sky2042:
> text changed to informative instead of prescriptive

That looks pretty reasonable to me.

GlazerMann:
@Sky2042 thank you.  I commented on the talk page too.  We shall see.

PR #1385, opened 2019-02-26
Better error handling

GlazerMann:
Report an error message for semi-recoverable conditions, instead of dieing in trigger_error().  

Instead of calling  trigger_error() for fatal problems call new report_error() which prints the error if not TRAVIS and then that calls trigger_error() to die -- and give error in TRAVIS.

This pull came about because blocked bot gives no error messages when trigger_error() is called.


GlazerMann:
@ms609 I just pulled it on the Wikipedia Servers and it works great.

PR #1389, opened 2019-02-27
Treat url and chapter-url at same time better

GlazerMann:
Fully do the chapter-url then the url.  Old code would not drop stuff if it was used.  Also, did not simplify URLS.

@ms609 I included a test!

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -3536,7 +3536,21 @@ final class Template {
     }
   }
 
-    
+  protected function unhide($par) {
+   foreach ($this->param as $p) {
+       if ($p->param == $par . '-hidden-CITATION_BOT_PLACEHOLDER') {
+         $p->param = $par;
+       }
+     } 
+  }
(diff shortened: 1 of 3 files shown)


Template.php
```diff
@@ -862,6 +862,10 @@ public function get_identifiers_from_url($url_sent = NULL) {
           // If no URL or website, nothing to worth with.
           return FALSE;
         }
+    } elseif (preg_match('~^' . MAGIC_STRING . '(\S+)$~', $url_sent, $matches)) {
```
ms609: Should we add a space after `chapter-url` and `url` to ensure that the match doesn't gobble up anything after a URL, such as a comment?  (Not sure whether this could happen in practice, but might not hurt to play it safe.)

GlazerMann: Space added

GlazerMann:
@ms609 space added

PR #1416, opened 2019-03-03
COOL: Travis supports PHP 7.3 now

GlazerMann:
FILTER_FLAG_HOST_REQUIRED actually is meaningless in PHP 5.6 and 7.*

String functions are getting more picky about passing them strings instead of numbers.

7.3 capitalization is better with &

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ language: php
 matrix:
   include:
     - stage: test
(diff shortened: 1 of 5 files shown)


ms609:
Presumably we're testing on Travis with the same version of PHP that's installed on (/that we're using on) Toolserver?

PR #1430, opened 2019-03-08
Deal better with new parameters

GlazerMann:
1. If you got unlucky and there was a comment in the first parameter, then it got copied to new ones
2. Save it, so that all new added parameters are added the same why irrespective of forgetting parameters
3. Do some extra checking for sanity


Template.php
```diff
@@ -3478,15 +3478,16 @@ public function set($par, $val) {
...
     }
+    $p = new Parameter();
     if (isset($this->param[0])) {
-      $p = new Parameter();
       // Use second param as a template if present, in case first pair 
-      // is last1 = Smith | first1 = J.\n
-      $p->parse_text($this->param[isset($this->param[1]) ? 1 : 0]->parsed_text()); 
+      // is last1 = Smith | first1 = J.\m
+      $example = $this->param[isset($this->param[1]) ? 1 : 0]->parsed_text();
+      if (stripos($example, 'CITATION_BOT_PLACEHOLDER') !== FALSE) $example = '| param = val';
```
ms609: Would 
```
$example = preg_replace(CITATION_BOT_PLACEHOLDER_BLAH, '', $example);
```
do a better job of keeping whatever format has already been used?  (Or could that just make things worse..?)

GlazerMann: Done.  Look at please.

GlazerMann:
Went crazy paranoid too.  Exactly one equals sign.  At least one character on each side of it too.

PR #1435, opened 2019-03-10
Remove fast mode

GlazerMann:
No-one seems to ever use this.  Non-slow is used on all interfaces to the bot, other than webpage.

Secondly, we do zotero, which is more slow than bibcodes ever were


ms609:
Is one solution to only query Zotero in $SLOW_MODE?

PR #1439, opened 2019-03-10
Clean up researchgate titles

GlazerMann:


Template.php
```diff
@@ -2916,6 +2916,12 @@ public function tidy_parameter($param) {
         case 'url':
           if (preg_match("~^https?://(?:www\.|)researchgate\.net/[^\s]*publication/([0-9]+)_*~i", $this->get($param), $matches)) {
               $this->set($param, 'https://www.researchgate.net/publication/' . $matches[1]);
+              if (preg_match('~^(.+){{!}} Request PDF$~', trim($this->get('title')), $match)) {
```
ms609: Do `{}`s need escaping?
```suggestion
              if (preg_match('~^(.+)\{\{!\}\} Request PDF$~', trim($this->get('title')), $match)) {
```

(Is it worth including a test that would catch this?)

Template.php
```diff
@@ -2916,6 +2916,12 @@ public function tidy_parameter($param) {
...
+                 $this->set('title', trim($match[1]));
+              }
+              if (preg_match('~^\(PDF\)(.+)$~', trim($this->get('title')), $match)) {
```
ms609: Is case important?
```suggestion
              if (preg_match('~^\(PDF\)(.+)$~i', trim($this->get('title')), $match)) {
```

tests/phpunit/TemplateTest.php
```diff
@@ -1657,6 +1657,15 @@ public function testTrimResearchGateETC() {
     $this->assertEquals('https://www.google.com/search?hl=en&safe=off&q=%22west+coast+hotel+co.+v.+parrish%22+(site%3Anewsweek.com+OR+site%3Apost-gazette.com+OR+site%3Ausatoday.com+OR+site%3Awashingtonpost.com+OR+site%3Atime.com+OR+site%3Areuters.com+OR+site%3Aeconomist.com+OR+site%3Amiamiherald.com+OR+site%3Alatimes.com+OR+site%3Asfgate.com+OR+site%3Achicagotribune.com+OR+site%3Anytimes.com+OR+site%3Awsj.com+OR+site%3Ausnews.com+OR+site%3Amsnbc.com+OR+site%3Anj.com+OR+site%3Atheatlantic.com)', $prepared->get('url'));
   }
  
+  public function testCleanRGTitles() {
+    $text = '{{cite journal|url=http://researchgate.net/publication/320041870_yup|title=Hello {{!}} Request PDF}}';
```
ms609: I suppose I'm wondering how this {{!}} gets into the title – is the bot converting a `|` when it finds one, or is this the work of some other bot?  If the former, perhaps a more useful test would be one where we give the bot a URL from which it extracts the title "Hello | Request PDF", and ends up with "Hello"?

PR #1477, opened 2019-03-15
Do not mess up html entities that end in semicolon

GlazerMann:


ms609:
Should probably add a test before merging?

PR #1481, opened 2019-03-15
trailing periods are not significant

GlazerMann:


ms609:
I wonder whether there's a case for a separate function `remove_irrelevant_dissimilarities` that could be applied to both strings?
Reducing code duplication could avoid the introduction of mistakes in the future, e.g. where one string is treated differently to the other.

PR #1501, opened 2019-03-17
Need to catch correct exception

GlazerMann:
That exception type does not get thrown by the library we use


GlazerMann:
@ms609 I did a test where I passed an invalid token and verified my belief

PR #1571, opened 2019-04-02
setup.php - create and split off from expandfns.php

GlazerMann:
I think you have to change the launch mechanism on the tool server once you upload this


GlazerMann:
@ms609 split off per your request

PR #1572, opened 2019-04-02
dx.doi.org sometimes returns type=book

GlazerMann:


ms609:
Should we add a test that triggers this behaviour?

PR #1574, opened 2019-04-03
list what files are changed for easier error understanding 

GlazerMann:
@ms609 What did I do that keeps breaking GIT???


ms609:
Interestingly I can run gitpull.php fine on the server, just not via the web... (?)

PR #1583, opened 2019-04-05
convert <em> to <i> on incoming text

GlazerMann:

### DIFF
--- a/expandFns.php
+++ b/expandFns.php
@@ -92,11 +92,10 @@ function wikify_external_text($title) {
   }
   $title = preg_replace('~[\*]$~', '', $title);
   $title = title_capitalization($title, TRUE);
-  
-  
+
   $htmlBraces  = array("&lt;", "&gt;");
   $angleBraces = array("<", ">");
-  $title = str_ireplace($htmlBraces, $angleBraces, $title)
+  $title = str_ireplace($htmlBraces, $angleBraces, $title);
 
   $originalTags = array('<title>', '</title>', 'From the Cover: ');
   $wikiTags = array('','','');
@@ -109,7 +108,7 @@ function wikify_external_text($title) {
       function ($matches) {return ("'" . $matches[1] . "'");},
       $title);
     $title = preg_replace_callback('~(?:<Emphasis Type="Bold">)([^<]+)(?:</Emphasis>)~iu',
-      function ($matches) {return ("<b>" . $matches[1] . "</b>");},
+      function ($matches) {return ("'''" . $matches[1] . "'''");},
       $title);
     $title = preg_replace_callback('~(?:<em>)([^<]+)(?:</em>)~iu',
       function ($matches) {return ("'" . $matches[1] . "'");},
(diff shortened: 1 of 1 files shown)


expandFns.php
```diff
@@ -95,10 +95,13 @@ function wikify_external_text($title) {
...
       $title);
   $title = preg_replace_callback('~(?:<Emphasis Type="Bold">)([^<]+)(?:</Emphasis>)~iu',
-      function ($matches) {return ("<b>" . $matches[1]. "</b>");},
+      function ($matches) {return ("<b>" . $matches[1] . "</b>");},
+      $title);
+  $title = preg_replace_callback('~(?:<em>)([^<]+)(?:</em>)~iu',
+      function ($matches) {return ("<i>" . $matches[1] . "</i>");},
       $title);
   
   $originalTags = array("<i>","</i>", '<title>', '</title>',"From the Cover: ");
```
ms609: Seems like we're using two approaches to solve the same problem here.

To be consistent, we should either (i) cut out the middleman and use `''` in the preg_replacement; or (ii) use preg_replacement rather than str_ireplacement for `<i>` and `</i>`, to ensure that we're only replacing matched tags.

expandFns.php
```diff
@@ -92,24 +92,33 @@ function wikify_external_text($title) {
...
+
+  $originalTags = array('<title>', '</title>', 'From the Cover: ');
+  $wikiTags = array('','','');
+  $title = str_ireplace($originalTags, $wikiTags, $title);
+
+  $title_orig = '';
+  while ($title != $title_orig) {
+    $title_orig = $title;  // Might have to do more than once.   The following do not allow < within the inner match since the end tag is the same :-( and they might nest or who knows what
+    $title = preg_replace_callback('~(?:<Emphasis Type="Italic">)([^<]+)(?:</Emphasis>)~iu',
+      function ($matches) {return ("'" . $matches[1] . "'");},
```
ms609: Two ''s for italics?
```suggestion
      function ($matches) {return ("''" . $matches[1] . "''");},
```

expandFns.php
```diff
@@ -92,24 +92,33 @@ function wikify_external_text($title) {
...
+      $title);
+    $title = preg_replace_callback('~(?:<Emphasis Type="Bold">)([^<]+)(?:</Emphasis>)~iu',
+      function ($matches) {return ("'''" . $matches[1] . "'''");},
+      $title);
+    $title = preg_replace_callback('~(?:<em>)([^<]+)(?:</em>)~iu',
+      function ($matches) {return ("'" . $matches[1] . "'");},
```
ms609:
```suggestion
      function ($matches) {return ("''" . $matches[1] . "''");},
```

expandFns.php
```diff
@@ -92,24 +92,33 @@ function wikify_external_text($title) {
...
+      $title);
+    $title = preg_replace_callback('~(?:<i>)([^<]+)(?:</i>)~iu',
+      function ($matches) {return ("'" . $matches[1] . "'");},
```
ms609:
```suggestion
      function ($matches) {return ("''" . $matches[1] . "''");},
```

PR #1587, opened 2019-04-06
Implicit acronyms

GlazerMann:
This actually should solve a lot of the capitalization additions.


expandFns.php
```diff
@@ -161,6 +161,17 @@ function title_capitalization($in, $caps_after_punctuation) {
     // ALL CAPS to Title Case
     $new_case = mb_convert_case($new_case, MB_CASE_TITLE, "UTF-8");
   }
+
+  // Implicit acronyms
+  $new_case = ' ' . $new_case . ' ';
+  $new_case = preg_replace_callback("~ [b-df-hj-np-tv-xz]{3,}(?= )~ui",
```
ms609: We could use `\s`, or possibly better still `\W` would match examples like the second half of `B/Gcvs`?
```suggestion
  $new_case = preg_replace_callback("~\W[b-df-hj-np-tv-xz]{3,}(?=\W)~ui", 
```

tests/phpunit/constantsTest.php
```diff
@@ -42,6 +42,24 @@ public function testConstantsDefined() {
     $this->assertEquals($text, $expanded->parsed_text());
   }
   
+  public function testImplicitConstants() {
+    // Consonants
+    $this->assertEquals('X', title_capitalization('x', TRUE));
+    $this->assertEquals('Xz', title_capitalization('xz', TRUE));
+    $this->assertEquals('XZZ', title_capitalization('xzz', TRUE));
```
ms609: Combining texts also allows us to check that sequences work correctly
```suggestion
    $this->assertEquals('XZZ BBBB/EEE', title_capitalization('xzz bbbb/eee', TRUE));
```

GlazerMann:
Cancelled tests since I need to shorten list of implicit acronyms that are not needed anymore

PR #1589, opened 2019-04-07
Only add title-link if a large part of title text

GlazerMann:
Otherwise it is almost always some thing like Journal of Organic [[Chemistry]]


josve05a:
Does this de-link "The [[Journal fo Science]]"? That should be kept, imo.

GlazerMann:
If the link is more than 60% of the title then it stays.

PR #1608, opened 2019-04-12
Fix volume=issue

GlazerMann:
If the citation has a DOI, and the doi agrees with one of them, then change the other one.

Even a test case.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -3248,11 +3248,8 @@ final class Template {
         $orig_data = trim((string) $this->get('volume'));
         $possible_issue = trim((string) @$crossRef->issue);
         $possible_volume = trim((string) @$crossRef->volume);
-        echo "$possible_issue\n";
-        echo "$possible_volume\n";
-        echo "$orig_data\n";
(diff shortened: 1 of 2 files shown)


Template.php
```diff
@@ -3241,6 +3241,27 @@ public function final_tidy() {
...
+        ($this->get('volume') == $this->get_without_comments_and_placeholders('volume'))) { // No comments to flag problems
+        $crossRef = query_crossref($this->get('doi'));
+        $orig_data = trim((string) $this->get('volume'));
+        $possible_issue = trim((string) @$crossRef->issue);
+        $possible_volume = trim((string) @$crossRef->volume);
+        if ($possible_issue != $possible_volume) { // They don't match
+          if ((strpos($possible_issue, '-') > 0 || (integer) $possible_issue > 1) && (integer) $possible_volume > 0) { // Legit data
+            if ($possible_issue == $orig_data) {
+              $this->set('volume', $possible_volume);
+              report_warning('Citation had volume and issue the same.  Changing volume.');
```
ms609: is `report_warning` the most appropriate function here?

PR #1616, opened 2019-04-19
Update capitalization.php

GlazerMann:
fix all consonant “word”


GlazerMann:
@ms609 This goes the other way.  It undoes the auto-capitalizaiton.

PR #1624, opened 2019-04-24
User based authentification - clean pull

GlazerMann:
Needs to upgrade the exising OAuth tokens so that they can also get user tokens at https://meta.wikimedia.org/wiki/Special:OAuthConsumerRegistration logged in as Citation Bot Account.

Then we get to debug on the development branch. You will also have to specify a return URL: I chose authenticate.php   

Unknown: Do we need a new edit token for editing of a file. ie. do we need to grab a new edit token for each and every file in a category?

Once we can do that, we should tag our edits (this is can be setup to automatically do this based upon our oauth keys) https://en.wikipedia.org/wiki/Special:Tags


davidbarratt:
so it looks like this still requires a `PHP_OAUTH_ACCESS_TOKEN` in my env, but if if I remove that line in `WikipediaBot.php` I just get a blank white page rather than (I assume) being redirected to the OAuth login.

davidbarratt:
I can fix this, but I don't want to break anything that is currently working on your end, since I assume there's an assuming in the code that I'm not understanding. 

I'll continue to dig into this and see if I can understand what's going on. :)

GlazerMann:
Reverse merge done

GlazerMann:
@davidbarrat I got the reverse merge done.

ms609:
I've proposed a new key for Travis, and will update the values there once the proposal is approved.
If I've set up everything correctly, then I'll propose an update to the production keys too.

pppery:
@ms609 Unless I'm mistaken, you need to make a request [on meta-wiki](https://meta.wikimedia.org/wiki/SRM) for your consumer to be approved.

davidbarratt:
also, you can start using the key while you are waiting, but only with the user that requested the key.

PR #1741, opened 2019-06-08
Turn pubmed zotero back on

GlazerMann:


GlazerMann:
@ms609 this test fails with a zotero server internal error.  Does the server need restarted?

PR #1747, opened 2019-06-10
Do not throw away Unpaywall OAI-PMH matches

nemobis:
Since this blacklist was added, Unpaywall/OAdoi has improved accuracy.
OAI-PMH author and title match also gets higher priority compared to other
methods like trusting whatever the metadata claims about DOI relationships.

### DIFF
--- a/Template.php
+++ b/Template.php
@@ -1943,7 +1943,7 @@ final class Template {
         }
         if (@$best_location->evidence == 'oa repository (via OAI-PMH title and first author match)') {
           // false positives are too common
-          report_warning("Unpaywall found an OA match on a repository via OAI-PMH" . echoable($doi));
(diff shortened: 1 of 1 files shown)


GlazerMann:
How sure are you about this?

nemobis:
> How sure are you about this?

I've manually sampled hundreds if not thousands of Unpaywall API responses in recent months and I don't see a correlation between OAI-PMH match and such false positives.

The specific issue of ad4595713485324e6940659cc1f74609e7d934e3 was a link to an abstract PDF which I've now reported at https://github.com/Impactstory/oadoi/issues/121 . We could also just blacklist it for `"journal_name": "Cochrane Database of Systematic Reviews"`: its specific problem is that it has a sister publication which duplicates every item, AFAICS.

Issue #1769, opened 2019-06-13
Combine requests to AdsAbs


ms609:
To reduce the number of calls that we are making to AdsAbs, the maintainers have suggested that we check that we've made maximal use of boolean search operators.

For example, in `Template.php::expand_by_adsabs`, could we combine what is at present two searches into a single search that looks something like "title = XYZ OR (journal = X AND volume = 123)"?

I've had a stab at combining identifiers in #1771, but are there any other opportunities to combine queries?

GlazerMann:
I cleaned it up and fixed it up.

PR #1771, opened 2019-06-13
Adsabs identifiers

ms609:
Contributes to #1769 

- Search for results that match any available identifier

To do: 
- Handle cases where multiple conflicting identifiers exist, and `$result->numFound > 1`.


ms609:
This looks like it should be okay to merge: Travis is encountering a couple of errors but these do not seem to be related to the present PR.

PR #1778, opened 2019-06-14
User Verification

GlazerMann:


davidbarratt:
I don't think this is a good idea, this would allow someone to put in any valid username to bypass their own block. :/

davidbarratt:
This also doesn't take partial blocks into account.

PR #1780, opened 2019-06-14
User based authentification - clean pull (fixes)

davidbarratt:
Making some fixes to the original patch so it works locally. Also removed the Citation Bot authentication token since it is not necessary.


GlazerMann:
@davidbarratt is this ready to try or what all is the next step to be done?

davidbarratt:
This is ready to try out! There is a small UX improvement I would like to make, but I can make that a seperate PR. :)

davidbarratt:
There we go... now it will redirect the user and then redirect them back with the query rather than just a message.

PR #1793, opened 2019-06-17
Add exception message

davidbarratt:


davidbarratt:
I'm not sure what is wrong with https://tools.wmflabs.org/citations/authenticate.php so adding additional details to figure it out.

PR #1975, opened 2019-07-21
Generic proxy removal code

GlazerMann:


GlazerMann:
What about urlsecode?????

PR #2122, opened 2019-08-28
Try to increase concurrency by half

nemobis:
The bot is currently impossibly slow to use for single pages while it is
successfully managing two large categories at once.
https://en.wikipedia.org/w/index.php?title=Special:Contributions&dir=prev&offset=20190828081954&limit=500&target=Citation+bot


nemobis:
GlazerMann, 28/08/19 15:49:
> Is it possible to specify different concurrency for different php files?

Unlikely.

nemobis:
GlazerMann, 28/08/19 15:49:
> Is it possible to specify different concurrency for different php files?

The docs are at
https://redmine.lighttpd.net/projects/1/wiki/Docs_ModFastCGI

PR #2176, opened 2019-10-05
Cookies

GlazerMann:


authenticate.php
```diff
@@ -1,4 +1,5 @@
 <?php
+@session_set_cookie_params(365*24*3600);
```
davidbarratt: Assuming that `.user.ini` gets run when the web server starts (you could use `phpinfo()` to verify the settings are correct) I don't think this is necessary.

.user.ini
```diff
@@ -1,5 +1,5 @@
 ini_set('session.cookie_httponly', 1); // Prevents javascript XSS attacks aimed to steal the session ID
 ini_set('session.use_only_cookies', 1); // Session ID cannot be passed through URLs, so people cannont share/bookmark them
 ini_set('session.cookie_secure', 1); // Uses a secure connection (HTTPS) if possible
-ini_set('session.gc_maxlifetime', (7 * 24 * 3600)); //One week duration
```
davidbarratt: This file is not in ini format.... are we sure it's being run when the webserver is started?

davidbarratt: See the examples here: https://www.php.net/manual/en/configuration.file.php

PR #2177, opened 2019-10-05
Rename .user.ini to set_session.php

GlazerMann:


davidbarratt:
It looks like Toolforge will allow the ini settings to be set:
https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web/Lighttpd#PHP
but it needs to be in the proper format (and the tool needs to be restarted any time they change).

davidbarratt:
Some ini settings cannot be set at runtime, so doing it properly might be better.

PR #2877, opened 2020-05-18
Wikipedia Proxy

GlazerMann:


ms609:
That was quick work!
I wonder whether there are other proxies that we ought to be able to detect, using a library of known proxies or some more general regexp.  I'd hoped that the original feature request might include suggestions as to how such a filter might be designed...

Samwalton9:
Thanks for the quick upgrade!

@ms609 I'll see if we can pull together a broader list of proxies :)

Samwalton9:
Something that needs to be considered here is that, at least in the case of EZProxy, any `.` in the URL will be replaced with `-`, even if there is already a dash in the URL. Reverse engineering that isn't as straightforward, since you need to do something like evaluate whether replacing `-` with `.` stops the URL from resolving. That should be an edge case, but could cause some problems.

GlazerMann:
That’s why I only allow two and three dashes in the hostname.   Since this pull request is closed and cannot be modified.  I suggest discussion of proxies be done on the Wikipedia talk page.  This is a nice addition to the existing proxy code.

PR #3414, opened 2020-08-25
Prevent undefined variable errors from $the_example_user

kaldari:
This prevents an undefined variable error being thrown every time someone loads the index page.


kaldari:
@GlazerMann - Test failure seems to be unrelated to the actual code change.

PR #3675, opened 2020-11-07
Merge over GitHub Actions code - probably need to set it up too somehow and remove TRAVIS-CI

GlazerMann:


ms609:
Removed branch protection so you can commit when ready, without awaiting Travis.

PR #3732, opened 2021-05-15
Add option to use the Citoid API in Zotero format

nemobis:
The only difference is that it takes GET requests rather than POST,
and the URL is passed in the URL rather than POST data.


tests/phpunit/TemplateTest.php
```diff
@@ -1351,23 +1351,25 @@ public function testOpenAccessLookup7() : void {
...
-   $this->assertSame('10.1145/358589.358596', $template->get2('doi'));
+   $template->get_unpaywall_url($template->get2('doi'), '');
+   $this->assertNotNull($template->get2('url'));
   }
-
+  
+  // Test Unpaywall OA URL does not get added when doi-access=free
   public function testUnPaywall2() : void {
-   $text = "{{cite journal|doi=10.1145/358589.358596|doi-access=free}}";
+   $text = "{{cite journal|doi=10.1007/s11075-018-0599-7}}";
```
nemobis, line 1370: doi:10.1007/s11075-018-0599-7 is actually a case which should find an URL from http://arxiv.org/abs/1804.09042 , but I suppose that's added as parameter rather than as URL. I'm not actually sure it's that important to test that doi-access=free citations are left alone.

nemobis:
Verbose phpunit says things like
```
>Remedial work to clean up templates</span>F
<span class='subsubitem'>  >Could not resolve URL https://www.nytimes.com/2018/06/11/technology/net-neutrality-repeal.html</span>F
<span class='subsubitem'>  >Could not resolve URL https://www.ncbi.nlm.nih.gov/books/NBK24662/</span>F
<span class='phase'>
```

nemobis:
I have no idea how `requires_zotero()` is supposed to work, but it's not working, and we don't really need it when the tests are not disabled, so I'm just emptying it.

PR #3978, opened 2022-05-08
Better session code

GlazerMann:


GlazerMann:
Needs caps and then tested 
In fact check true/false caps in all existing code

PR #4139, opened 2023-02-06
Update bad_data.php

sounderbruce:
Removing The Washington Post from being converted from work= to newspaper=, a constant complaint in the bugs thread

Fixes # .

-
-


GlazerMann:
Will fix in a better way

PR #4234, opened 2023-12-06
Ability to specify PubMed API Key and email via enrvironment variable

maximmasiutin:
This pull request adds ability to specify PubMed API Key and email via enrvironment variable.

See `env.php.example`:

```
  @putenv('NLM_APIKEY=xxxxx');       // API key for *nlm.nih.gov, can be obtained at https://account.ncbi.nlm.nih.gov/settings/
  @putenv('NLM_EMAIL=xxxxx');        // Email address for *nlm.nih.gov
```

If no PubMed email is specified in the environment variable (or `xxxxx` is specified), the default value from constants.php will be used. If no API Key is specified (or `xxxxx` is specified, or a value shorter than 8 characters is specified, then this parameter will not be passed to PubMed as was before. The API key allows higher request rate. You can get one from https://account.ncbi.nlm.nih.gov/settings/

Also, if you specify the PubMed API key, then it will not try one more time after sleeping 1 or 3 seconds on failure, because there is no reason to do so when using an API key; request count is not an issue; moreover, when using an API key, you can contact them to increase rate even higher.

I tried with my API Key and it works pretty well!

### DIFF
--- a/.github/workflows/php-static2.yml
+++ b/.github/workflows/php-static2.yml
@@ -44,7 +44,7 @@ jobs:
       run: composer install --prefer-dist  -vvv --no-progress
 
     - name: PHP phpstan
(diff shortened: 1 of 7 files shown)


GlazerMann:
You can ignore the size test.  Template.php is already too bid.  Which means that the github search does not index it.

PR #4246, opened 2023-12-07
Fixed character conversion tests that failed on server, added more verbose output for the test

maximmasiutin:


GlazerMann:
On another positive note, the wikipedia server supports PHP 8.2, so I was able to move onto that.

PR #4247, opened 2023-12-07
When a DOI check or a HDL check returned with a timeout, do not try gain multiple time this DOI

maximmasiutin:
When a DOI check or a HDL check returned with a timeout, do not try gain multiple time this DOI.

A DOI check can return one of the following results:
* TRUE: DOI OK
* FALSE: DOI Fail
* NULL: check again later

In the past, DOIs such as 10.3746/pnf.2017.22.2.67 took about a minute to check to pass all the timeouts, still that URL returned NULL and was rechecked over and over again. Now, with this pull request, a time is measured low long did it take to check a DOI. If it took more than a timeout (BOT_CONNECTION_TIMEOUT or BOT_HTTP_TIMEOUT, whichever larger), than we consider that this check returned FALSE (Fail) rather than NULL, so we would not recheck such slow DOIs. Only DOIs which returned quickly will be rechecked.

With this pull request, the bot no longer spends hours on DOIs such as 10.3746/pnf.2017.22.2.67. This DOI presents on the page Cholesterol, so you can check it there.

### DIFF
--- a/expandFns.php
+++ b/expandFns.php
@@ -19,23 +19,15 @@ function doi_active(string $doi) : ?bool {
   // For really long category runs
   if (count($cache_bad) > MAX_CACHE_SIZE) $cache_bad = [];
   if (count($cache_good) > MAX_CACHE_SIZE) $cache_good = [];
-  $start_time = time();
   $works = doi_works($doi);
   if ($works === NULL) {
-    // if little time passed, we will recheck again, otherwise mark as fail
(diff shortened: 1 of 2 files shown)


GlazerMann:
Indeed, this is always a problem with NULL/BAD/GOOD.

maximmasiutin:
> I have personally complained about hundreds of NULLs and got _some_ of them fixed.

That was just a null because of timeouts.

GlazerMann:
I merged everything but the timeout.  Will do that after tests run.

PR #4255, opened 2023-12-08
CrossRef api use: see https://api.crossref.org/swagger-ui/index.html

### LINKED ISSUE #4252: Specifying email in URI and User-Agent for CrossRef
@GlazerMann 

CrossRef API is described at https://api.crossref.org/swagger-ui/index.html

To use CrossRef with lower rate limit, we should specify the same email address in both request URI and the user-agent. However, we already have a user-agent line that contains an email address specified. 

https://github.com/CrossRef/rest-api-doc#good-manners--more-reliable-service

Moreover, we should probably specify email once, rather than multiple addresses for PubMed, CrossRef, etc... How do you think it is the best way to make configuration for proper user agent and URI for CrossRef?

maximmasiutin:
Fixes #4252

With this fix, the server returns the following headers, confirming the use of the proper API:
```
< x-ratelimit-limit: 50
< x-ratelimit-interval: 1s
< x-api-pool: polite
 ```

P.S. We can later use functions to return contexts and user agent strings, to allow email addresses configurable via environment variables.


maximmasiutin:
@GlazerMann - how can we fix that size error? Should we increase the limits or split the files? I can think on how split the files if you wish. Just tell me the path to follow. Anyway, did you like this pull request on Crossref API?

GlazerMann:
>  how can we fix that size error? Should we increase the limits or split the files? 
We cannot do either.  The bad file is Template.php and the whole template has to be in one file.  We can slowly pull bits out.

GlazerMann:
Merging bit by bit

maximmasiutin:
I guess we should implement one-time diagnostic message. When the bot invokes CrossRef API, it should check whether the API returned "x-api-pool: polite". If it returned, the bot should report information such as "CrossRef API confirmed that that the citations bot is using the API with self identification which provides higher request rate". If there is no such header returned, the bot should notify a warning "CrossRef API reported that the citations bot is using anonymous API that provides lowest quality of service". This way, we will see how it goes.

PR #4290, opened 2023-12-20
Refactor ADS curl API calls

maximmasiutin:
Refactored ADS curl API calls to make them less monolithic, easier to understand and expand. It also allows to close curl handle before starting analyzing the response. There are the following functions added:
1) curlGetResponse - retrieves http status code, headers and body from the curl handle, gives fully documented exceptions in case of errors.
2) formatUrlResponse - returns a formatted error (warning) message when request to a particular URL failed, accepts an URL, optionally two error messages as strings. This message can be logged as warning or error or raised as an exception.
3) Bibcode_Response_Processing parameters are updated; removed the curl handle ($ch) and instead added http_response_code, header, and body. Therefore, we can start processing the bibcode response when curl handle is already closed. The http_response_code integer and header + body strings are returned by curlGetResponse.
4) made curl close handle in "finally" block to prevent handle leak.

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -372,10 +372,10 @@ function adsabs_api(array $ids, array &$templates, string $identifier) : bool { 
       $body = '';
       curlGetResponse($adsabs_url, $return, $ch, $http_response_code, $header, $body);
       unset($ch);
+      $response = Bibcode_Response_Processing($adsabs_url, $http_response_code, $header, $body);
   } catch (Exception $e) {
       return FALSE;
   }
-  $response = Bibcode_Response_Processing($adsabs_url, $http_response_code, $header, $body);
 
   if (!isset($response->docs)) return TRUE;
 
@@ -1518,6 +1518,7 @@ function query_adsabs(string $options) : object {
       $header = ''; 
       $body = '';
       curlGetResponse($adsabs_url, $return, $ch, $http_response_code, $header, $body);
+      unset($ch);
       $response = Bibcode_Response_Processing($adsabs_url, $http_response_code, $header, $body);
     } catch (Exception $e) {
       return (object) array('numFound' => 0);
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -1278,7 +1278,7 @@ function curlGetResponse(string $url, string $return, CurlHandle $ch, int &$http
     // @codeCoverageIgnoreStart
     $errorStr = curl_error($ch);
     $errnoInt = curl_errno($ch);
-    throw new Exception('Curl error from adsabs website' . $errorStr), $errnoInt);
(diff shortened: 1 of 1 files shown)


GlazerMann:
curl_close() does nothing in PHP 8.  curl object automatically de-construct when unset or go out of scope.

GlazerMann:
I see no value in including the URL in the error message.  What does that add?

GlazerMann:
I have merged in the latest main branch changes

GlazerMann:
I have migrated most of this into the main code branch, other than the splitting of the function into two parts.  I have added logging to the unusual errors.

maximmasiutin:
@GlazerMann 

Please consider accepting this pull request rather than closing it unaccepted for two reasons:

1. The proposed code first extracts the body and header from the curl handle and closes the handle (step 1) and then parses them (step 2). These two steps are clearly delineated and easy to understand.
2. The proposed code has consistent error message formatting via a single function, otherwise there are plethora of ad-hoc formatting:
```
      throw new Exception('Curl error from adsabs website: ' . $errorStr, $errnoInt);

    if ($http_response_code === 0 || $header_length === 0) throw new Exception('Size of zero from adsabs website');

        bot_debug_log("ADSABS website returned a stack trace - URL was:  " . $adsabs_url);
        throw new Exception("ADSABS website returned a stack trace" . "\n - URL was:  " . $adsabs_url,

         throw new Exception(((isset($decoded->error->msg)) ? $decoded->error->msg : $decoded->error) . "\n - URL was:  " . $adsabs_url,

        throw new Exception('Too many requests', $http_response_code);

      bot_debug_log("Could not decode ADSABS API response:\n" . $body . "\nURL was:  " . $adsabs_url);
      throw new Exception("Could not decode API response:\n" . $body, 5000);  // @codeCoverageIgnore

      throw new Exception("Could not decode AdsAbs response", 5000);        // @codeCoverageIgnore

      report_warning(sprintf("API Error in query_adsabs: %s", echoable($e->getMessage())));

      report_warning(sprintf("Error %d in query_adsabs: %s", $e->getCode(), echoable($e->getMessage())));
```

These refactoring that I propose do not change how the program work, but improve the design of existing code making it easier to develop in the future. You wrote that URL or host are irrelevant for logging, but as you see, sometimes URL is reported (`ADSABS website returned a stack trace - URL was: `)

I was considering the following development. When the AdsAbs ran out of request credits, to implement an option to do one of the following:
1. remove the big_give_up() and small_give_up() which count the number of requests, but this count is irrelevant.
2. Analyze the response on when the API will be available again (using the header returned)
3. Do one of threee options depending on configuration:
 - skip API calls until the API will be available again based on analysis in p.2
 - wait (sleep) until the API will be available again
 - exit the bot if we run out of credits.

PR #4304, opened 2023-12-30
Clear ratelimit status reporting (refactoring)

maximmasiutin:


maximmasiutin:
Thank you very much! Let me also make a pull request to sleep on excess of AdsApi calls, I tested that during past few days.

PR #4310, opened 2023-12-30
A configurable action: what to do when AdsAbsAPI limit reached

maximmasiutin:
A configurable action: what to do when AdsAbsAPI limit reached: exit, sleep or ignore (default). Some changes are not actually changes but whitespace/tab changes for consistency. We should be consistent in using tabs or spaces: select a way and follow it, rather then have somewhere tabs and somewhere spaces.

This configurable action is set in the env.php file as an environment variable.

By default, old behaviour is implemented (ignore).

In the current implementation, it sleeps exact number of seconds until the AdsAbsAPI limit resets, plus one second. However, after sleep, it does not retry current request. This may be an small item for the todo list (I suggested a refactoring/split to several functions to implement it easier but you didn't like the split). But still failing just one request in 5000 (daily limit) is much better than failing and failing over and over again as in the current version.

Can you please help with this one at https://github.com/ms609/citation-bot/actions/runs/7366768549/job/20049330782
```
ERROR: UnevaluatedCode - apiFunctions.php:1305:11 - Expressions after return/throw/continue (see https://psalm.dev/084)
          exit(1);
```


maximmasiutin:
@GlazerMann: I should have probably not inserted "exit()" because the checker probably sees that the report error function exits anyway?

GlazerMann:
> I should have probably not inserted "exit()"
Yes, the error function is declared to have a return value of "never".

PR #4323, opened 2024-01-07
Better edit summary for singulars in "Add" and "Alter"

maximmasiutin:
Better edit summary for singulars, i.e, when only 1 (one) change or alteration was done, the semicolon is not used, and a different, friendlier form of the verb is used.

For example, instead of former use of "Add: s2cid." it will be "Added s2cid."

But when the number in the list is higher than one, the old algorithm is used, i.e., "Add: pages, issue, volume."

### DIFF
--- a/Page.php
+++ b/Page.php
@@ -528,16 +528,22 @@ class Page {
 
   public function edit_summary() : string {
     $auto_summary = "";
-    if (count($this->modifications["changeonly"]) !== 0) {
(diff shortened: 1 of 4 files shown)


maximmasiutin:
Oops, I should have updated the test cases as well, because they rely on older spelling. Let me update them. Sorry that I forgot to do that first time.

PR #4463, opened 2024-03-04
Use GET requests instead of POST wherever possible to make the requests cacheable via a proxy

maximmasiutin:
Use GET requests instead of POST wherever possible to make the requests cacheable via a proxy. This allows using a local proxy server (even on the same operating system where the bot is running) to increase throughput and reduce the chance of hitting API limits.

For the adsabs API calls when only a single bibcode identifier is requested, we can use GET instead of POST. The API url is [https://ui.adsabs.harvard.edu/help/api/api-docs.html#get-/search/query] -- the current implementation sometimes requests multiple ids but sometimes one id. Calls for multiple IDs are kept supported the old way via POST, but if we need just one ID, we use cacheable GET. 

Pubmed (NCBI) requests will always be called via GET.


maximmasiutin:
I run on various real scenarios, it works OK.

I've figured out that unit tests gave `UnusedVariable` on `$bibcode = "";` -- I just set the value for safety, this line is not actually needed. I could remove it, but didn't want to run the full suite of tests again:
```
ERROR: UnusedVariable - apiFunctions.php:380:5 - $bibcode is never referenced or the value is not used (see https://psalm.dev/077)
    $bibcode = "";
```

maximmasiutin:
@GlazerMann can you please remove this line `$bibcode = "";` that is before `foreach` because if I do that it will run through the full test suite again.

GlazerMann:
Some of these do not seem to work

maximmasiutin:
> Some of these do not seem to work

I understand that some of the modifications I made that switch from POST to GET, may not seem to function as expected; still, the potential benefits of using GET over POST for certain operations are very big, because GET requests are cacheable via proxies, while POST requests are not, meaning that repeated GET requests can be served faster, improving the overall performance and responsiveness of the bot when using a proxy, to avoid duplicate requests. While it’s true that some of these changes may not work perfectly at the moment, I believe that with further refinement and testing, we can overcome these initial hurdles :-)

maximmasiutin:
> Some of these do not seem to work

Still, in the future, could you please try to use GET wherever possible rather than POST, for better caching?

GlazerMann:
I think there is some oddities with the url-escaping

maximmasiutin:
Maybe we can make a few reliable test cases? What do you think?

maximmasiutin:
@GlazerMann -- let us handle this open issue so it could be closed. What action is required from me to get this closed?

PR #4698, opened 2024-11-23
Added ruwiki for bot operation

In1quity:


GlazerMann:
I looks like ru does not use CS1/2 with English words

PR #4807, opened 2025-10-17
Bad title: Rate limit reached

redalert2fan:
Bad title: Rate limit reached


redalert2fan:
From here: https://en.wikipedia.org/wiki/User_talk:Citation_bot#Bad_title:_%22Rate_limit_reached%22

PR #4826, opened 2025-11-28
disable/skip tests causing CI failures

### LINKED ISSUE #4825: CI is broken
Looking at commits from this year, continuous integration (CI) tests for this repo are almost never all green. The master branch should be kept green at all times, so that when a PR/commit fails a test, we know that it's that PR/commit that is the problem.

Quick solution:
- mark any flaky/failing tests as skipped, so that we can properly pay attention to the working tests

Thorough solution:
- take the time to fix each of the failing tests

NovemLinguae:
Why
- get master branch green, so that we can see when a PR/commit causes a test failure

What
- disable codecov. is a frequent source of test failures, and maintainers merge patches failing this test anyway, so it is not being enforced
- mark 14 tests from "Bot Full Test Suite" as skipped. this suppresses 14 failures and 0 errors.

Fixes #4825

### DIFF
--- a/tests/phpunit/TemplateTest.php
+++ b/tests/phpunit/TemplateTest.php
@@ -1505,6 +1505,8 @@ final class TemplateTest extends testBaseClass {
     }
 
     public function testOpenAccessLookup1(): void {
+        // TODO: fix this test
+        $this->markTestSkipped();
         $text = '{{cite journal|doi=10.1136/bmj.327.7429.1459}}';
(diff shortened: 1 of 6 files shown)


GlazerMann:
Please add a // TODO comment to all lines about it needing fixed.  Also,  double check the latest Master action to see if we have any of them fixed.  Many seems to be that S2 is not working.

GlazerMann:
too many changes to sync up.

PR #4828, opened 2025-11-29
Prevent the bot from mangling arXiV citations

sohomdatta1:
TLDR, some of the recent issues like [1] and [2] are caused by what appears to be a API change on ArXiV's end. The current implementation assumes that the ArXiV URL data are ordered in the order they were requested (in the `$ids` variable in line 270 in apiFunctions.php).

This patch adds a step that cleans up the ordering of the data ArXiV returns so that before we start fixing templates, the `$ids` array has the same ordering as the data returned by arXiV.

[1]: https://en.wikipedia.org/wiki/User_talk:Citation_bot#CitationBot_merging_info_from_two_completely_different_references
[2]: https://en.wikipedia.org/wiki/User_talk:Citation_bot#ArXiv_references_getting_completely_messed_up

### DIFF
--- a/apiFunctions.php
+++ b/apiFunctions.php
@@ -293,8 +293,24 @@ function arxiv_api(array $ids, array &$templates): void {  // Pointer to save me
         return;
     }
 
-    $this_template = current($templates); // advance at end of foreach loop
+    // Arxiv currently does not order the data recieved according to id_list. This is causing CitationBot to mix up
+    // which Arxiv ID is associated with which citation. As a result, we first perform a sorting pass to make sure we
+    // order the arxiv data based on our id_list so that we have a 1 to 1 ordering of both.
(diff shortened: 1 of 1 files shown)


GlazerMann:
This is actually buggy, I have created a new pull to fix it 
https://github.com/ms609/citation-bot/pull/4858
Your gets out of order if there is an item missing, and then does horrible things

GlazerMann:
Also. this:
https://github.com/ms609/citation-bot/commit/e4d7af266c56fd7394ec6ce05c621082cdbda6aa

GlazerMann:
You have run into the most annoying thing about this bot.  Data from exterior databases and data from wikipedia is often not what you expect.  And, even when it is for decades, it might suddenly stop being.  Defensive programming.

PR #4854, opened 2025-11-29
Replace mb_ucfirst_bot with mb_ucfirst

GlazerMann:


redalert2fan:
Seems like the Bot Full Test Suite / build (pull_request) is stuck at 1403 / 1626 ( 86%), it has been there for over one hour now on Full test with PHP.

PR #4871, opened 2025-11-30
Delete dependabot.yml attempt 2

redalert2fan:
Delete dependabot.yml attempt 2

to recreate it next


GlazerMann:
good luck, but I think the problem is that we use *.yml, not composer.json files

PR #4874, opened 2025-11-30
README: add architecture diagram

NovemLinguae:
I spent a couple hours studying the bot tonight. These are my notes.


GlazerMann:
Two minute time limit is not for jobs, because it keeps getting reset, but it should never actually be hit.  It is just there in case PHP goes of the rails.  Note that remote access times do not count towards this limit.

PR #4964, opened 2025-12-03
temporarily disable psalm

### LINKED ISSUE #4948: PSALM is goofy
https://github.com/php/php-src/issues/17866

NovemLinguae:
Why
- psalm is broken due to an upstream bug

What
- temporarily disable psalm. we can try re-enabling it in the future. the .yml files will remain in the repo in a "disabled" folder

fixes #4948 PSALM is goofy


GlazerMann:
I have figured out how to disable JIT

Issue #4956, opened 2025-12-02
linter is not enforcing indentation


NovemLinguae:
I see a linter (overtrue/phplint) in CI and in the composer.json dev section, but I am encountering a lot of un-linted code in this repository. The most obvious thing is that indentation is inconsistent across files. For example, in Template.php, the file uses one space as indentation. In one of the test files, it uses 4 spaces for indentation, but some of the lines have 5 spaces, which is an off-by-one error that a developer probably didn't notice due to linting not working properly. (By the way, this is one of the upsides of tabs... you don't get things like the 5 spaces issue. But I digress.)

Anyway, this ticket is to investigate what's going on here. Either fix the linter, or add a rule for indentation. Auto-fix everything, then have the linter enforce the rule going forward.

GlazerMann:
Template.php is SO big that switching to 4 spaces will actually push it over the GitHub size limit for files being indexed for search.

GlazerMann:
Template.php shrunk

GlazerMann:
As I split up the code into files, they have all sorts of bad indentation.  Needs fixed.

NovemLinguae:
Yeah. Want to reply back with a list of non mb_ functions that you'd like to blacklist? We can probably figure out how to get PHP CodeSniffer to forbid those functions.

NovemLinguae:
We WANT mb_ functions, correct? So we'd want to blacklist NON mb_ functions, correct? So we'd want a list of non mb_ functions to blacklist, I think.

GlazerMann:
Here is the black list
convert_case
ereg
ereg_replace
ltrim
rtrim
str_pad
str_split
stripos
strlen
strpos
strrev
strrpos
strtolower
strtoupper
substr
trim
ucfirst
ucwords

PR #4996, opened 2025-12-07
Template.php: fix indentation

NovemLinguae:
Why
- Template.php has odd indentation (1 space) compared to the rest of the repo (4 spaces)
- We should start setting style lint rules in a rules file (.phpcs.xml), doing autofixes using phpcbf, and we should keep things we've fixed from creeping back into the repo by adding the style linter to CI.

What
- Run PHP CodeSniffer autofixes on Template.php, using only indentation rules.
- I used 4 spaces as the standard, although I personally prefer 1 tab
- Also add PHP CodeSniffer to CI. I copied the .yml file from #4992, but removed the cron job. I don't see why a style linter would need to run weekly. I also simplified the composer install command to `composer install -vvv --no-progress`. I also renamed the file to .phpcs.xml, which is the file name MediaWiki code uses for this.
- Remove CI check looking for large files. Per comments below, sounds like we don't need to enforce this anymore.

In the future, we can slowly autofix more files and add more rules.

TODO before merging:
- manually test that we didn't break the bot with our Template.php changes
- ✅ let the CI job PHP CodeSniffer complete and make sure it passes

If you want to recreate what I did on localhost...
- check out the patch
- `docker exec -it citation-bot-php-1 /bin/bash` - opens a shell inside the container
- `composer update` - to install PHP CodeSniffer
- `./vendor/bin/phpcbf --standard=ruleset.xml -d memory_limit=1000M` - runs the autofixes. increased memory limit needed because Template.php is so big. lol

`phpcs` does read only scans. `phpcbf` does auto fixing. They are installed by the same package, squizlabs/php_codesniffer.

Related #4956 (but doesn't close out that ticket yet, since we still need to autofix more files)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -12,7 +12,6 @@ to test.
 All code is run through several tests.  The primary test is a suite of example pages and citation templates.  There are a variety of static code analysis tests that look for common errors.
 The security tainted data tests make sure that all "untrusted input" (data from wikipedia pages) is output wrapped with the echoable() function: this is not done primarily for security, but for proper output formatting.
 The GitHub provided CodeQL test suite is also enabled, but that just checks the one JavaScript file.
-Because large files are not indexed by GitHub, there is a test to check for large files: the test will suggest LFS on failure, but we do not use LFS. The DOI database files currently fail, but they are not really code.
 
 ## Submitting changes
 


NovemLinguae:
Template.php failed the large size test. Guess it's too big even with moving 1000 lines to another file.

Want me to do tabs instead? That'd be less characters than spaces.

GlazerMann:
No problem with file size.  Just use spaces for now.

GlazerMann:
A bunch of the API files have wrong spacing since they just got pulled out of Template.php

PR #5120, opened 2025-12-18
phpcs: fix linter errors with less than 10 occurrences

NovemLinguae:
Why
- fix linter errors with low #'s of impacted lines, because this is pretty easy to write a patch for and pretty easy to review
- that way we can 1) reduce the # of excluded rules and 2) get these rules turned back on to protect against new code style issues creeping into the codebase

What
- manually fix 21 rules (see .phpcs.xml removed lines for a list of fixed rules)
- manually fix several docblocks

Notes
- the idea behind docblocks is to be a comment that attaches itself to a class, function, or variable, and provides a tooltip whenever and wherever this class/function/variable is hovered over by your mouse
image

- however, computers are computers, so the syntax has to be exactly right for this to work
- docblock:
```
/**
 * the start line must be forwardslash asterisk asterisk.
 * every line after the first must be space asterisk space.
 * the end line must be space asterisk forwardslash.
 * it must be touching the class, function, or variable you want it to document, with no blank lines
 * or `//` comments in between them.
 * if the docblock is bigger than 1 line, leave the first and last line blank
 */
```

- generic comment (not a docblock):
```
/* Start it with 1 asterisk. Put it wherever. Type
as many lines as you want */
```

### DIFF
--- a/src/includes/Template.php
+++ b/src/includes/Template.php
@@ -197,7 +197,7 @@ final class Template
         }
     }
 
-    // Re-assemble parsed template into string
+    /** Re-assemble parsed template into string */
     public function parsed_text(): string
     {
         if ($this->has(mb_strtolower('CITATION_BOT_PLACEHOLDER_BARE_URL'))) {


NovemLinguae:
The DesignSecurity failure is a false test failure due to some kind of timeout when downloading the GitHub Action. I doubt it's related to this patch.

PR #5165, opened 2025-12-23
Try slightly newer version out

GlazerMann:


NovemLinguae:
Will bundle upgrading to 48.0.0 into my next phpcs patch, which I'll post in a couple hours.

PR #5259, opened 2025-12-26
Fix test file naming inconsistency - rename templatePart2Test to TemplatePart2Test

redalert2fan:
This pull request makes a minor update to the test suite by renaming the `templatePart2Test` class to `TemplatePart2Test` for consistency with naming conventions.

### DIFF
--- a/tests/phpunit/includes/TemplatePart2Test.php
+++ b/tests/phpunit/includes/TemplatePart2Test.php
@@ -7,7 +7,7 @@ declare(strict_types=1);
 
 require_once __DIR__ . '/../../testBaseClass.php';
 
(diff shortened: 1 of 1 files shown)


GlazerMann:
The Lower t makes the test run later, and it is shorter than other ones.  It is a quick way to optimize the tests.

redalert2fan:
Ah okay I get why you changed it to this now, well in that case this pull can be considered not needed, but it would be probably good to add a note since it isn't intuitive per se to  have one test with different capitalization from others.

GlazerMann:
I think the only way to get a different order is to specify every test file in the xml config file.  I will change the names so that 4 runs last and add a note

PR #5260, opened 2025-12-26
Make test names more human reasonable and add note to explain

GlazerMann:


redalert2fan:
Consider this human happy - copilot should also understand the note

PR #5507, opened 2026-03-30
Fix testUrlReferencesAAAAA

redalert2fan:
output from API was changed, causing test to fail


redalert2fan:
of course they changed the type of dash used as well...

PR #5543, opened 2026-04-28
Remove bogus `|issue=Volume N, YEAR` injected by Visual Editor for Annual Reviews citations

redalert2fan:
This pull request improves the handling of bogus "issue" parameter values in citations for Annual Review journals, specifically cleaning up cases where Visual Editor inserts values like "Volume N, YEAR" into the `issue` field. The changes ensure that such values are either moved to the `volume` field or removed, depending on whether `volume` is already set, and only when the year appears plausible.
**Bug fix and data cleanup for Annual Review citations:**

* Updated `tidy_parameter` in `Template.php` to detect and handle `issue` values of the form "Volume N, YEAR" (or similar), moving the volume number to the `volume` field if it is not already set, or removing the bogus `issue` if `volume` is present. This only applies if the year is in the 1900s or 2000s, and the match is case-insensitive and allows for an optional comma.

**Testing improvements:**

* Added a suite of PHPUnit tests in `templatePart4Test.php` to verify correct cleanup of Annual Review bogus `issue` values, including tests for case insensitivity, optional comma, and ensuring only plausible years trigger the cleanup.

### DIFF
--- a/src/includes/Template.php
+++ b/src/includes/Template.php
@@ -5706,7 +5706,7 @@ final class Template
                         $this->forget($param);
                         return;
                     }
-                    if (preg_match('~^Volume\s+(\d+),?\s*\d{4}$~i', $value, $matches)) {
+                    if (preg_match('~^Volume\s+(\d+),?\s*(19|20)\d{2}$~i', $value, $matches)) {
                         if ($this->blank('volume')) {
                             $this->rename($param, 'volume', $matches[1]);
                         } else {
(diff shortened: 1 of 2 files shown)


NovemLinguae:
The first post in this PR states that it adds tests, but there are no tests added in the files changed tab.

Were the tests accidentally removed?

redalert2fan:
Uuuh good catch, where did the test go indeed.

redalert2fan:
https://github.com/ms609/citation-bot/pull/5543/changes/1745b2372610acc3187072cc42be884d173ebba4 
Merge branch 'master' into copilot/cleanup-remove-issue-from-citations overwrote them with test from one of my other PR's.

redalert2fan:
Restored lost changes

PR #5599, opened 2026-06-13
Fix labeler config for v5/v6 format and update to v6

redalert2fan:


GlazerMann:
Please check my fixes

PR #5601, opened 2026-06-13
Fix typos and grammar across codebase

redalert2fan:


redalert2fan:
seems like labeler also works now

PR #5605, opened 2026-06-14
Remove unsupported open-access parameter per bug report

redalert2fan:
When encountering |open-access=yes, the bot was Levenshtein-matching it to |osti-access=yes, which creates CS1/CS2 errors (yes is not a valid value for osti-access, and it requires matching |osti=).

Fixes: https://en.wikipedia.org/wiki/User_talk:Citation_bot#improper_replacement_of_unsupported_parameter


redalert2fan:
1) ConstantsTest::testWhiteListNotBlacklisted
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'{{citation | open-access = Z123Z }}'
+'{{citation }}'

Is a valid test failure since open-acces is now removed by the code - test should be adjusted

redalert2fan:
fixed

PR #5653, opened 2026-07-15
Fix/title Strip non-printable characters (U+FFFC and control characters) from Zotero/Citoid-sourced titles.

redalert2fan:
## Summary

Strip non-printable characters (U+FFFC and control characters) from Zotero/Citoid-sourced titles.

## Problem

When a citation URL contains percent-encoded non-printable characters such as `%ef%bf%bc` (U+FFFC Object Replacement Character), the decoded character can propagate into the page's HTML `<title>` tag. The Citoid/Zotero API returns this character in the `title` field, and the bot stores it verbatim with no sanitization.

For example, a URL like `https://example.com/page%ef%bf%bc/` results in a citation title containing the literal U+FFFC character.

## Fix

In `process_zotero_response()`, strip non-printable control characters (U+0000–U+0008, U+000B, U+000C, U+000E–U+001F, U+007F–U+009F) and U+FFFC from `$result->title` once on the object immediately after the data is parsed, before it reaches any of the three code paths that consume it.

## Changes

- **`src/includes/api/APIzotero.php`**: Add `safe_preg_replace` cleanup of non-printable characters from `$result->title`
- **`tests/phpunit/includes/api/zoteroTest.php`**: Add two tests verifying U+FFFC and control characters are stripped

### DIFF
--- a/src/includes/api/APIzotero.php
+++ b/src/includes/api/APIzotero.php
@@ -397,6 +397,10 @@ final class Zotero {
         unset($result->nameOfAct);
         unset($result->language);
         unset($result->source);
+        // Strip non-printable characters from title (e.g. U+FFFC from URL percent-encoding)
+        if (isset($result->title)) {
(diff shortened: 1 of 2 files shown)


redalert2fan:
test fail not related to this pull - flaky test

PR #5666, opened 2026-07-19
DEV_USERS: allow trusted developers to exceed normal category limit

redalert2fan:
Adds a `DEV_USERS` constant in `category.php` allowing trusted developers
to bypass the normal category page limit of 12 when running category
requests via POST.

**Changes:**

- New `DEV_USERS` constant defines the whitelist (`AManWithNoPlan`,
  `Redalert2fan`)
- If the user is in `DEV_USERS` and `MAX_PAGES_OVERRIDE` is not already
  set (from a GET-whitelisted category), the override is set to 1,000,000
- Edit summary appends `| Developer - max category limit override` for
  dev runs vs `| Whitelisted category` for GET-whitelisted category runs

**Edit summary combinations:**

| Scenario | Suffix |
|---|---|
| Normal user, ≤12 pages | *(none)* |
| Normal user, >12 pages | Cancelled |
| GET-whitelisted category, ≤250K | `\| Whitelisted category` |
| DEV_USER, ≤12 pages | *(none)* |
| DEV_USER, 13–250K pages | `\| Developer - max category limit override` |
| Any, >250K pages | Cancelled |

### DIFF
--- a/src/category.php
+++ b/src/category.php
@@ -20,6 +20,7 @@ const GET_IS_OKAY = [
     'CS1 errors: extra text: pages',
     'CS1 errors: extra text: volume',
     'CS1 errors: invisible characters',
+    'CS1 maint: article number as page number',
     'CS1 maint: bibcode',
     'CS1 maint: date format',
     'CS1 maint: extra punctuation',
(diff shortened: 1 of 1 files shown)


src/category.php
redalert2fan, line 1: adding category for related request

PR #5688, opened 2026-07-25
try out php 8.6

GlazerMann:


GlazerMann:
The only thing that failed was some of the tools!

PR #5690, opened 2026-07-26
Fix Docker image security scanning by ignoring unfixed CVEs and applying OS patches

redalert2fan:
Reduces 1300+ spurious Trivy alerts from Docker image scanning:

1. **Add `ignore-unfixed: true` to Trivy step** — The vast majority of
   alerts were for Perl and kernel CVEs that Debian has never backported
   fixes for. The filesystem scan already used this flag; the Docker
   image scan was missing it.

2. **Apply `apt-get upgrade`** — Fixes 14 real CVEs in packages Debian has
   patched (libblkid, libxml2, Unbound, etc.).

3. **Restructure Dockerfile layers** — Merges 8 RUN commands into 3 to
   reduce layer count (minor optimization, neutral on CVE count).


redalert2fan:
that should fix some of the 1.3K code scanning alerts I caused by enabling the docker build check

GlazerMann:
Layer count reduction is good idea.  I like this pull.

GlazerMann:
Remove draft status when yoh want commit

redalert2fan:
Just one more commit needed and it will hopefully be okay again

redalert2fan:
Completed

PR #5714, opened 2026-08-07
Guard against exhausted template list in arxiv_api

redalert2fan:
current()/next() return Template|false, so add a defensive guard before using \, resolving 24 PHPStan level 7 errors.


redalert2fan:
failed test are  not related to this patch, will open a new patch for that

redalert2fan:
#5716

PR #5716, opened 2026-08-07
Fix PHP_CodeSniffer command injection in *blame reports

redalert2fan:
Bump mediawiki/mediawiki-codesniffer from ^51.0.0 to ^52.0.0, which pins squizlabs/php_codesniffer 3.13.6. This fixes a command injection vulnerability in the Gitblame, Hgblame, and Svnblame reports when processing untrusted files (patched in PHP_CodeSniffer v3.13.6/v4.0.2).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require-dev": {
         "brianium/paratest": "^7",
         "designsecurity/progpilot": "^1",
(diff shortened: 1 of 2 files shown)


redalert2fan:
bot full test suite fail not related to this patch

PR #5717, opened 2026-08-07
Fix/doi access free pr

redalert2fan:
## Summary

- Avoid adding duplicate `doi-access=free` parameters.
- Upgrade existing non-free DOI access values when the DOI is known to be free.
- Add regression tests for both cases.
- Use `mb_substr_count()` in tests to satisfy PHPCS.

## Test Plan

- Focused DOI access tests pass: 5 tests, 9 assertions.
- Changed files are limited to `Template.php` and `templatePart4Test.php`.


redalert2fan:
test fail unrelated to this patch, should be fixed by #5716

PR #5905, opened 2026-08-30
Throttle

GlazerMann:
Does not seem to work when deployed


GlazerMann:
This test worked
for i in $(seq 1 60); do
    curl -s -o /dev/null -w '%{http_code}\n' \
      'https://citations.toolforge.org/generate_template.php?doi=10.1038/nature12373' &
done
wait

PR #5833, opened 2026-08-26
Cs1 conformance fixes

redalert2fan:
## Summary

Fixes three CS1 citation-conformance issues found by a systematic audit of the bot code against the **Help:CS1 errors** categories (`docs/CS1-error-audit.md`). Each change closes a case where the bot could emit output that makes `Module:Citation/CS1` raise an error, and also fixes two Wikipedia talk-page bug reports (42, and the series-in-title part of 20).

## Changes

- **`src/includes/TextTools.php`** — add `isbn_valid()`: ISBN-10 (mod-11, `X`=10) and ISBN-13 (mod-10) check-digit validation, with CS1's `978`/`979` prefix and `9790` (ISMN-reserved) checks, and normalization of lowercase `x` / non-ASCII dashes.
- **`src/includes/Template.php`**
  - `add_if_new('isbn')` now rejects ISBNs with an invalid check digit, bad length, or invalid prefix (`Check |isbn= value`), logging a `report_inaction` when rejected.
  - Cite web → cite book conversion: when `work=` names a book series (`is_book_series('work')`) and `series=` is blank/equivalent, it is routed into `series=` instead of `title=` (`|work= ignored`, `|chapter= ignored`); an existing different `series=` is never clobbered.
  - `tidy()` now removes an orphaned URL access parameter (`url-access`, `chapter-url-access`, etc.) when its base URL parameter is genuinely absent — checking the hyphenated, non-hyphenated alias, and uppercase (`URL`) base forms (`|<param>-access= requires |<param>=`, talk-page 42).
- **`progpilot.json`** — whitelist a header-injection false positive in `src/authenticate.php` (guarded by `is_valid_local_return_path()` + a whitespace check; introduced by upstream's recent changes).
- **Tests** — `isbn_valid` unit tests (check digits, prefixes, ISMN, lowercase `x`), `add_if_new('isbn')` gate, cite web→book series routing (both paths + no-series-clobber), orphaned/alias/uppercase access-param handling; three fixtures updated to valid ISBNs.

## Notes on trade-offs

- **ISBN gate:** invalid-check-digit ISBNs from APIs (CrossRef/Zotero/Google Books) are now *dropped* rather than added. This aligns with the community consensus that "no ISBN is better than an incorrect ISBN" (talk-page 19), and is logged via `report_inaction`.
[...]

PR #5911, opened 2026-08-31
Add CS1 self-validation harness and identifier format validation

redalert2fan:
## Summary

Tier 2 of the CS1-conformance plan: the **self-validation harness** (Phase 0, built before the fixes so every change is verified against it), the **first fix batch** (identifier-format validation on add), and the **CI job** that runs the harness as a standing regression gate. Rebased on the current master.

## 1. CS1 self-validation harness (`tools/cs1_harness.php`)

Runs 36 citation templates through the bot's real expansion machinery (fast + slow modes) and flags any expanded output that would trigger a CS1 error, using the encodable rules from the audit (`docs/CS1-error-audit.md`).

- **Checker:** 16 rule families — title non-empty, no orphaned `trans-*`/`*-access` (all alias forms), ISBN check-digit/prefix, archive-url/archive-date coupling (incl. `archiveurl`/`archivedate`/`accessdate` aliases), `access-date` requires url, no bare external links, periodical params only in periodical-capable templates, `cite web` requires url, identifier-required templates, `cite document` requires publisher, coupled `doi-broken-date`/`format`/`pmc-embargo-date`, url+title-link conflict, and identifier-format checks (arxiv/pmid/pmc/bibcode/biorxiv/medrxiv).
- **Matrix:** 25 must-pass cases (violation → exit 1) + 11 documented **known-gap** cases (current CS1 violations the bot leaves in place — bad ISBN in input, url-less cite web, empty title, malformed identifiers in input, etc.). Gaps are reported but don't fail the run; they become must-pass as later tiers land. A gap that stops violating prints `RESOLVED`, so gaps can't rot into false greens.
- **Design notes:** `SLOW_MODE` is a compile-time define, so the harness runs as two processes (`--slow`). It forces `CI=1` and pre-defines `SLOW_MODE`, with a PHPUnit-style error handler that honors `@` suppression (the PHP 8 `INSUPPRESSIBLE_LEVELS` mask) and escalates only the `report_*` severities — so `report_error()` becomes a per-case failure instead of `exit(0)`. No credentials needed; the matrix uses fabricated identifiers/titles that APIs cannot match, so pass/fail is reproducible when upstream APIs respond normally (verified arXiv returns an empty feed for the fake IDs used).

## 2. Identifier-format validation on add (Tier 2)

New validators in `src/includes/TextTools.php`: `arxiv_id_valid`, `pmid_valid`, `pmc_valid`, `rxiv_id_valid`, `bibcode_valid` — gated in `Template::add_if_new` with `report_inaction`, mirroring the merged ISBN gate (PR #5833). Malformed identifiers from sources are dropped instead of written.

- Validators are deliberately **at least as permissive as the bot's own URL-extraction regexes** (a stricter CS1-faithful version broke existing tests) — defense-in-depth, not strict CS1 fidelity, documented in comments.
- Coverage: arXiv old/new forms (+versions), PMID/PMC 1–8 digits (PMC prefix normalized), bioRxiv/medRxiv legacy 6–8 digit + dated forms (incl. the newer `10.64898` prefix), bibcode length/year/charset (CS1's full body charset).
- The cite journal→cite biorxiv/medRxiv conversion path also validates the DOI before converting.

## 3. CI job (`.github/workflows/cs1-harness.yml`)

Runs the harness fast + slow on push/PR touching `src/**` or `tools/**`, plus a weekly schedule. Mirrors `phpstan.yml` (hash-pinned actions, PHP 8.4, `composer validate`, composer cache). Added to `test-suite.yml`'s `paths-ignore`.

## Tests & intent preservation

- **TDD:** new validator tests (5) + gate tests (4), watched fail before implementing.
- **Fixtures updated deliberately** (each reviewed for intent preservation):
  - `testPMCExpansion3`: 10-digit → **`PMC99999999`** — 8-digit (CS1-valid), genuinely non-existent (verified HTTP 404; the intermediate 7-digit `PMC9999999` turned out to be a **real article**).
  - `testMedRxivComprehensiveParameterFiltering`: malformed `10.64898/test123` → non-existent dated `10.64898/2099.12.31.999999` (the "real-looking" candidate was a **real bioRxiv article** — network dependence avoided).
  - `testId2Param3/4/5`: `astr.ph/1234.5678` (a malformed hybrid) → `zzzz/1234567` (valid old-style, non-existent archive → network-independent).
  - `testAddBogusBibcode_2` → renamed `testRejectBogusBibcode`, plus new `testPaddedShortBibcodeAccepted` (guards the still-live pad-to-19 path).
- **New regression guards** for behaviors the fixture changes masked: hybrid archive+new-style arXiv rejection, malformed-`10.64898` rejection, day/month range.

## Verification

- Harness: **25 pass / 11 gaps / exit 0** in both modes, deterministic across runs.
- Targeted identifier tests all green; full-suite CI is the authoritative gate.
- phpcs clean; phpstan unchanged (one pre-existing `setup.php` issue also on master).
- Rebased cleanly onto current master (69 commits ahead of the original base).

## Notes

- The `10.5555/` DOI handling change that fixed the `testTidy12` jstor flake arrived upstream during this patch's lifetime — that known-flake note is now moot.
- Known unrelated flakes that may appear in CI: `zoteroTest::testZoteroExpansion_*`, `unpaywallApiTest`, some pubmed tests — pre-existing network dependence.
- Internal planning/audit docs are intentionally kept out of this PR.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -90,6 +90,7 @@ ci:
       - .github/labeler.yml
       - .github/cffconvert-requirements.txt
       - tools/update_cffconvert_lock.sh
+      - tools/cs1_harness.php
 
 docker:
   - changed-files:
--- a/.github/workflows/cs1-harness.yml
+++ b/.github/workflows/cs1-harness.yml
@@ -1,0 +1,59 @@
+name: CS1 Harness
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'src/**'
+      - 'tools/**'
+      - 'composer.json'
+      - 'composer.lock'
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'src/**'
+      - 'tools/**'
+      - 'composer.json'
+      - 'composer.lock'
+  schedule:
+    - cron: '10 18 * * 1'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+    - name: Setup PHP
+      uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # 2.37.2
+      with:
+        php-version: '8.4'
+        coverage: none
+        extensions: mbstring, intl, curl, openssl, xml, sockets, pcntl
+
+    - name: Cache Composer dependencies
+      id: composer-cache
+      uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: vendor
+        key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-composer-
+
+    - name: Install dependencies
+      run: composer install -vvv --no-progress
+
+    - name: Run CS1 harness (fast mode)
+      run: php tools/cs1_harness.php
+
+    - name: Run CS1 harness (slow mode)
+      run: php tools/cs1_harness.php --slow
--- a/.github/workflows/cs1-harness.yml
+++ b/.github/workflows/cs1-harness.yml
@@ -49,6 +49,9 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-composer-
 
+    - name: Validate composer.json and composer.lock
+      run: composer validate
+
     - name: Install dependencies
       run: composer install -vvv --no-progress
 
--- a/.github/workflows/cs1-harness.yml
+++ b/.github/workflows/cs1-harness.yml
@@ -8,6 +8,7 @@ on:
       - 'tools/**'
       - 'composer.json'
       - 'composer.lock'
+      - '.github/workflows/cs1-harness.yml'
   pull_request:
     branches: [ master ]
     paths:
@@ -15,6 +16,7 @@ on:
       - 'tools/**'
       - 'composer.json'
       - 'composer.lock'
+      - '.github/workflows/cs1-harness.yml'
   schedule:
     - cron: '10 18 * * 1'
 
@@ -32,6 +34,8 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        persist-credentials: false
 
     - name: Setup PHP
       uses: shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240 # 2.37.2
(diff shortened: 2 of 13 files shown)


GlazerMann:
Here is what my paid for Ai thinks.  Some seem good and some seem meh

https://chatgpt.com/s/t_6a956188ee9c8191961e03adad99e51e

GlazerMann:
And GitHub seems to have comments too

redalert2fan:
> And GitHub seems to have comments too

theres about 20 of those open, this just adds a similair issue. am going to fix all of them in  a seperate patch

redalert2fan:
> Here is what my paid for Ai thinks. Some seem good and some seem meh
> 
> https://chatgpt.com/s/t_6a956188ee9c8191961e03adad99e51e

This patch is  just the start of a large implemtation, so some of the issues AI has are valid, but they are in my task plan for next patches already so they arent in this pull. Will take a closer look and mark and undraft when done with that.

redalert2fan:
Assessment
1. Harness is self-confirming (high priority)
Agreed. check_citation() calls the exact production validators the PR introduces (arxiv_id_valid, pmid_valid, pmc_valid, bibcode_valid, rxiv_id_valid), so a validator bug is invisible to the harness. That defeats the harness's purpose as a conformance oracle. The harness's CS1 checks should be encoded independently of the implementation (e.g. cs1_arxiv_valid() etc. following the actual Module:Citation/CS1 rules). We're treating this together with point 2 since they're the same architecture decision.

2. Separate "extractable" from "safe to emit into CS1"
Agreed, and it matches the plan. Our plan's wording for this tier is "mirrors the strict DOI validation already present" — so the intended design is: permissive recognition at extraction (URL/API), strict CS1 validation immediately before writing the parameter. The current patch relaxed the validators to match the URL-extraction regexes to keep existing tests green; that was a pragmatic call, but it means values like 1234.5678 (impossible month) or a padded 2019.............. (no letter at position five) can still be written into citations, which CS1 then flags. We're deciding whether to rework this properly now or land it as a follow-up; either way the two-level design is correct.

3. URL-alias handling in the harness
Agreed, genuine bug. ACCESS_PARAMS handles url/URL and chapter-url/chapterurl, but R9 (cite web requires url), R13 (format/chapter-format), and R15 (url + title-link) check only the canonical names. We'll fix by defining shared alias groups (URL_PARAMS = ['url','URL'], CHAPTER_URL_PARAMS = ['chapter-url','chapterurl']) and using has_any() everywhere, plus matrix cases for URL + access-date/archive-url/format and chapterurl + chapter-format.

4. Unexpectedly-resolved gaps should fail
Agreed. The current runner prints RESOLVED for a gap that stops violating but still exits 0 — which contradicts the "no false greens" intent. We'll treat it as an XPASS: count it separately and exit nonzero so a resolved gap must be explicitly converted from gap to pass (or confirmed intentional).

5. CI cleanup
Agreed. cs1-harness.yml should trigger on changes to itself (add it to both push and pull_request path lists), and the actions/checkout step should set persist-credentials: false.

Plan
3, 4, 5 — will fix in this PR (small, contained).

1, 2 — Continue the current permissive version first and track the strict-validation + independent-oracle work as a follow-up (the plan already has related follow-ups in Tiers 4–5).

Thanks again for the review

redalert2fan:
ready, but feel free to review

PR #5959, opened 2026-09-03
Always tag category runs as #UCB_Category and retire personal #UCB_ tags

redalert2fan:
- Category runs via the webform now always emit #UCB_Category (scope wins over caller), so they stay countable separately from single-page #UCB_webform runs in statistics.

- process_page no longer mints #UCB_Headbomb, #UCB_Smith609 or #UCB_arXiv from ?edit=
   Personal tags could have been spoofed by ayone appending ?edit= with the username to result in the #UCB tag being added
   #UCB_arXiv is not used anywhere
   They fall now through to #UCB_Other; the tags stay in KNOWN_UCB_TYPES with a deprecation note so historic edits still classify.


redalert2fan:
To celebrate the bot passing 6.1 million edits today, lets make sure our new statistic page accurately reflects the activation type.

PR #5969, opened 2026-09-04
Add big-run gate so single requests take priority over bulk runs

redalert2fan:
## Summary

Adds a global **big-run gate** so single requests (webform runs of ≤4 pages) take unconditional priority over big runs (category, linked-pages, and large webform lists). With 24 homogeneous PHP-FPM workers and no queue, priority is enforced as admission control on big runs.

### Behavior
- **Single** (≤4 pages): never gated.
- **Small big** (5–49 pages) / **Large big** (≥50 pages): nested concurrency pool — total ≤ 10 big runs, of which large ≤ 4.
- **Token bucket**: capacity 400, refill 4.0/s, charged at admission, non-negative, no refund. Cost = `min(400, ceil(pages × type_weight × size_weight))`.
- Type weights: category/linked/webform 1.5, automated_tools/template 0.5, toolbar/other 1.0, testing 0.0 (fully exempt).
- Size weights: small 1.0, large 1.5.
- Exempt: DEV_USERS, CLI, gadget, singles, testing.
- Deferred runs get a reason-aware busy page (big_full with active count, or tokens with a buffered wait estimate).
- Big runs show a responsibility reminder notice.

### Changes
- `src/includes/RequestRateLimit.php`: `big_run_try_acquire`/`big_run_release` + constants, in one locked state file (`big-run.json`), fail-open on storage errors.
- `src/includes/WebTools.php`: `gate_big_run`, `big_run_type_from_edit`, `big_run_gate_decision`, busy-page helpers; `edit_a_list_of_pages` gains a `run_type` parameter.
- Entry points pass the activation type (`category`, `webform_linked`, or mapped from `?edit=`).

### Testing
- RequestRateLimitTest + WebToolsTest additions (nested cap, token arithmetic, type×size costs, reason selection, 20% wait buffer, stale pruning, fail-open).
- New `BigRunGatePageTest` (separate-process; uopz tests run in CI where the extension is installed).
- phpcs/phpstan/psalm/psalm-taint clean; CS1 harness passes in fast + slow mode.


redalert2fan:
@GlazerMann let me know what you or the AI thinks so far about this patch

GlazerMann:
From one of the AI

Good, I have the branch. Let me compute the actual merge-base diff so I'm looking at exactly what the PR would introduce.I dug into the actual diff (cloned both branches and diffed `fork/big-run-gate` against the merge-base, since GitHub's API is rate-limited right now). Here's a substantive review, organized by concern.

## `RequestRateLimit.php` — the gate mechanics

**1. Lock contention causes unnecessary rejections.**
`big_run_try_acquire` uses `flock($handle, LOCK_EX | LOCK_NB)` and immediately gives up with `reason = 'retry_later'` if the lock is busy — even when slots and tokens are actually available. Since only *big* runs (rare, ≤10 concurrent) ever touch this lock, a single short bounded retry (e.g. 2–3 attempts with a few-millisecond sleep) before giving up would likely eliminate a class of false "come back later" pages without meaningfully risking worker pile-up.

**2. Token bucket can be emptied by a single request.**
A max-weight large run (`category`/`webform`, weight 1.5 × size 1.5 = 2.25/page) hits the 400-token cap at just ~178 pages. At a 4/s refill rate, that's **100 seconds** before any other big run can be admitted — independent of the concurrency-pool check. That may be intentional, but it's worth explicitly sanity-checking against real traffic post-deploy rather than shipping the constants as final; consider whether they should be tunable (env/config) rather than hardcoded, given this is a brand-new subsystem.

**3. Fail-open is a soft spot, not just a safety net.**
Every storage failure path (`base_directory === ''`, mkdir failure, fopen failure) admits the run unconditionally. That's a defensible choice consistent with the rest of this file's philosophy, but it does mean the entire gate degrades to a no-op if the state directory becomes unwritable (disk full, permissions drift, etc.) — worth being deliberate that this is monitored, not just logged once per worker process via `request_rate_limit_log_failure`.

## `WebTools.php` — gate integration

**4. Interaction with the existing `big_jobs_check_overused` per-user lock wastes tokens.**
In `edit_a_list_of_pages`, `gate_big_run()` runs *before* `big_jobs_check_overused()`. For a ≥50-page run, if a user already has one big job running, `gate_big_run` will admit the run (spending tokens, no refund by design) and register a release, and only *then* does `big_jobs_check_overused` reject it and `exit(0)`. The shutdown function frees the concurrency slot, but the spent tokens are gone for nothing. Worth swapping the order (cheap per-user check first) so the shared token budget isn't drained by requests that were always going to be rejected for an unrelated reason.

**5. Minor doc/code mismatch.**
`gate_big_run()`'s docblock says "callers must only invoke this when `HTML_OUTPUT` is true," but the function itself independently re-checks `!HTML_OUTPUT`. Not a bug — just slightly confusing since the comment implies a precondition the code doesn't actually rely on. Either drop the redundant check or soften the comment.

**6. `edit=testing` bypass — already closed, worth noting it's solid.**
I traced this carefully since it's the one place a user-controlled parameter feeds the gate: `big_run_type_from_edit()` explicitly refuses to map `?edit=testing` to the `'testing'` weight/exemption, with a comment explaining why. Good catch on the author's part (matches the "Close edit=testing gate bypass" commit) — I didn't find another way to reach the `testing` exemption from an HTTP entry point.

## Testing

**7. Test gap around the interaction in #4.**
`RequestRateLimitTest`/`WebToolsTest`/`BigRunGatePageTest` cover the gate in isolation well (nested pool, token math, pruning, fail-open, lock contention), but I didn't find a test exercising the `gate_big_run` → `big_jobs_check_overused` sequence for a ≥50-page run from the same user, which is exactly the scenario in #4.

## `progpilot.json`

**8. Line/column-pinned suppression is fragile by design (already flagged by the author, but worth reinforcing).**
The 692-line addition whitelists false positives keyed to exact source line/column numbers, and `AGENTS.md` now warns future contributors not to add lines above a certain point in `RequestRateLimit.php` or the suppression silently breaks. That's a real maintenance hazard for a security-scanning gate — any unrelated future edit to that file has a good chance of quietly reintroducing a "known" false positive as a real CI failure (or worse, quietly losing coverage). If the tool supports content/AST-anchored suppression instead of line/column, it'd be worth migrating rather than accumulating more of these.

---

Overall the design is solid and the code is careful (locking, fail-open, capped costs, non-negative token math, stale-entry pruning) — my suggestions are mostly about interaction with existing code (#4) and operational tuning (#1, #2) rather than correctness bugs. Want me to sketch the actual patch for the ordering fix in #4, since that one's the most concrete?

redalert2fan:
I will take a look at the suggestions, thanks

GlazerMann:
The AI give a pile of money to:

I reviewed the current head of PR #5969 (32c4a5f5…), including the implementation, entry points, tests, existing big_jobs behavior, PR discussion, and CI. The overall approach is sound, but I would request changes before merging. The previous review caught several good issues; I found two additional architectural/correctness concerns that I consider more important.

Overall assessment

The nested concurrency pool plus token bucket is a reasonable design for protecting a 24-worker PHP-FPM deployment. The implementation is generally careful: shared state is locked, costs are bounded, token math is conservative, the user-controlled edit=testing bypass has been explicitly closed, malformed state recovers safely, and test coverage is substantially better than I would normally expect for this sort of admission-control change.

The current PR is still a draft and mergeable, and the current head remains 32c4a5f5….

My findings, in priority order:

1. HIGH — The gate happens too late to fully protect single requests. In category.php, the code authenticates and calls WikipediaBot::category_members() before edit_a_list_of_pages() reaches gate_big_run(). Category enumeration is not trivial: category_members() requests up to 500 members at a time and follows cmcontinue until the entire category has been enumerated.   For linked-page jobs, get_links() and parsing similarly happen before admission.   This means a burst of, say, 24 category requests can still occupy all 24 PHP-FPM workers doing OAuth/API/category discovery before any of them reaches the 10-run gate. That directly weakens the main purpose of the PR. I would introduce a small preflight/discovery lease for inherently bulk entry points such as category and linked-pages. Acquire a provisional bulk slot before the remote enumeration, determine the actual filtered page count, then either release it for ≤4 pages or upgrade/charge the proper token cost. This is the biggest thing I would fix before deployment.
2. HIGH — A legitimate running job can be declared stale after five minutes. BIG_RUN_STALE_TIMEOUT_SECONDS is 300 and pruning is based solely on the original started_at.  But the page-processing loop can run much longer than five minutes, and WikipediaBot::fetch() calls set_time_limit(120) again for API operations, effectively refreshing the PHP execution timer during a long multipage run.  Once an admitted job reaches 300 seconds of wall-clock age, another acquisition will remove it from the state even though its worker can still be processing pages. The gate can then admit replacements and exceed both the 10-total and 4-large concurrency guarantees. I recommend storing last_seen_at and refreshing the lease from the existing per-page loop—near big_jobs_check_killed() is a natural place. Prune based on heartbeat age, not initial start time.
3. MEDIUM-HIGH — Shutdown release can silently fail and create a five-minute ghost job. big_run_release() uses LOCK_EX | LOCK_NB; if another process has the lock for even a few milliseconds, release simply returns. The entry then occupies capacity until stale pruning.  This is more troublesome to me than nonblocking acquisition. Acquisition can reasonably fail fast to protect FPM; release should make a much stronger effort to succeed. A bounded retry—for example a few attempts separated by a very short sleep—or even a normal blocking flock() for this extremely short critical section would substantially reduce ghost slots.
4. MEDIUM — The existing per-user big-job rejection occurs after tokens have already been spent. This is the strongest point from the earlier review. gate_big_run() admits and deducts tokens first, then big_jobs_check_overused() can discover that the same user already has a ≥50-page run and exit.  The shutdown callback releases the slot but deliberately does not refund tokens. The existing per-user mechanism clearly blocks a second large job for the same user.  Move that rejection/precheck ahead of token charging, or split big_jobs_check_overused() into a non-mutating availability check followed by acquisition.
5. MEDIUM — Deferred responses should have real HTTP overload semantics. big_run_try_acquire() calculates what its docblock calls Retry-After, but big_run_render_busy_page() only displays a warning and exits; it does not set Retry-After, Cache-Control: no-store, or a non-200 status.   This matters particularly because automated_tools is explicitly a supported gated request type. The repository already does this correctly for gadget rate limiting: it sends Retry-After, Cache-Control: no-store, and HTTP 429.  I would use a 503 Service Unavailable for global worker/capacity contention, probably also for the global token bucket, with Retry-After. If the maintainers interpret token exhaustion as client rate limiting, 429 is another defensible choice.
6. MEDIUM-LOW — Large-subpool wait calculations use the wrong oldest job. When a large request is rejected specifically because four large jobs are running, the code obtains big_run_oldest_started_at($entries) across all jobs.  An older small job can therefore determine the estimated slot wait even though that small job finishing would not free a large slot. This can also alter whether the returned reason is big_full or tokens. Have big_run_oldest_started_at() accept a tier filter, or separately calculate the blocking pool: all entries when total=10, large entries when large=4.
7. LOW — Keep fail-open, but monitor it. I agree with the existing choice not to take Citation Bot offline merely because the state file is unavailable. But mkdir/open/write failure makes the entire capacity mechanism ineffective.  Logging is useful, but this deserves an operational signal/metric because an unwritable temp directory could silently turn the new feature into a no-op.
8. LOW — Clean up documentation and wording. The PR description says the tests cover a 20% wait buffer, while the implementation and tests use 30% (×1.3).   Also, I would soften “single requests take unconditional priority.” DEV_USERS explicitly bypass the gate, and category/linked discovery currently occurs before admission. Something like “caps non-exempt bulk processing to preserve worker capacity for small requests” is more technically accurate.

What I would require before merge

I would make #1 and #2 merge blockers. They affect the central guarantee the PR is trying to establish. I would also strongly prefer #3 and #4 in the same PR because both are small changes with meaningful effects on state/accounting correctness.

The HTTP-status improvement and tier-specific wait calculation could reasonably be a follow-up if the maintainers want to control scope.

The testing strategy itself is good. The unit tests cover the nested cap, token math, corrupt state, lock contention, stale pruning, exemptions, and user-facing messages, and there is a separate-process integration test around the web gate.  The checks I inspected on the current head are green across the full test suite and the major static/security/build workflows. I would add regression tests specifically for heartbeat renewal, release under lock contention, per-user rejection without token expenditure, and preflight gating before category enumeration.

Recommendation

Request changes, but keep the architecture. I would not replace the token bucket/concurrency design. The main improvement is to turn its current “start timestamp + late admission” model into a small lease system:

preflight lease → discover/filter pages → finalize admission/token charge → periodically renew lease → reliably release.

That would make the implementation substantially closer to the guarantee described by the PR while preserving most of the existing code and tests.

GlazerMann:
I ask the AI to find internal inconsistencies and dubious assumptions.  I have found this question to be very valuable in addition to the open ended answer.

I found a second layer of concerns. Several are not ordinary coding bugs; they are places where the implementation, documentation, tests, and deployment model disagree or depend on assumptions that are not enforced.

Additional findings

1. HIGH — the “24 PHP-FPM workers” assumption needs verification before merge

The PR’s entire choice of BIG_RUN_MAX_TOTAL = 10 rests on there being 24 homogeneous PHP-FPM workers. The repository does not establish that. The documented production launch is:

webservice --backend=kubernetes php8.4 start

Current Toolforge documentation describes php8.4 as using a FastCGI helper process, not PHP-FPM.  Toolforge’s documented lighttpd PHP configuration shows two FastCGI backends with two PHP children each; absent a citations-specific out-of-repository configuration, that is dramatically different from 24 workers. 

This matters enormously. If there are only 4 effective PHP workers, allowing 10 simultaneous bulk jobs reserves exactly zero capacity for small requests.

I would make verification of the actual production process tree a merge blocker. Check the live citations service for:

* actual PHP worker count;
* $HOME/.lighttpd.conf;
* any service template or custom FastCGI settings;
* whether workers actually share /tmp;
* number of Kubernetes replicas.

Then either encode the worker count/reserved capacity in deployment configuration or make BIG_RUN_MAX_TOTAL configurable. A safer model is approximately:

max_bulk = actual_worker_capacity - reserved_interactive_capacity

rather than hard-coding 10 based on an undocumented 24.

2. HIGH — the token weights are partially controlled by the requester

This is more serious than the already-fixed edit=testing issue.

The code explicitly recognizes that ?edit= is user-controlled, so it refuses to honor testing. But it still honors:

* edit=template → weight 0.5
* edit=automated_tools → 0.5
* edit=toolbar → 1.0
* arbitrary unknown value → other → 1.0
* ordinary webform → 1.5

The tests explicitly codify that behavior.

So a user who would normally submit a 50-page webform run can simply claim edit=template:

* normal webform: ceil(50 × 1.5 × 1.5) = 113 tokens
* claimed template: ceil(50 × 0.5 × 1.5) = 38 tokens

That is nearly a 3× quota discount from an untrusted attribution field.

Even a made-up edit=whatever changes it from 1.5 to the 1.0 other weight.

The code correctly recognized that a user-controlled string must not grant the testing exemption, but the same trust-boundary reasoning should apply to all quota discounts.

Recommendation: attribution and resource policy should be separate concepts. Keep edit= for edit-summary/statistics attribution if desired, but derive the admission cost from trusted server-side context. If the source cannot be authenticated, charge the conservative/default cost.

3. HIGH — threshold splitting trivially defeats both tiers

The design assumes that a request’s page count meaningfully represents a workload. But the requester can split workloads.

For example:

* 5-page webform → gated
* 4 pages + 1 page → both ungated

Likewise:

* 50-page webform → large tier, large-subpool cap, about 113 category/webform tokens
* 49 pages + 1 page → 49-page small tier + an ungated request, about 74 tokens total, while completely avoiding the large-run subcap

This is particularly conspicuous because ordinary web requests have MAX_PAGES = 50.  Thus the ≥50 “large” tier for ordinary page-list web requests largely means exactly 50 pages.

The term “single request” is also misleading: ≤4 pages is not actually a single-page request.

If this mechanism is designed only for cooperative traffic shaping, this may be acceptable, but that assumption should be stated. If it is intended as robust resource protection, it needs some combination of per-user concurrency/rate accounting and a distinction between a true single-page interactive request and small batches.

4. HIGH — run_type has two incompatible meanings

This explains another inconsistency I mentioned while reviewing.

For page-list runs, run_type means caller/source:

webform, template, automated_tools, toolbar.

For category/linked runs it instead means workload shape:

category, webform_linked.

Those are two separate dimensions being represented by one variable.

The inconsistency becomes visible in categories. The existing code deliberately says edit=template and edit=automated_tools should override the normal category edit tag; tests explicitly verify that.  But category.php always charges the gate as 'category'.

So an operation can simultaneously be:

* recorded as #UCB_template,
* described as a template activation,
* documented as having weight 0.5,
* but actually charged category weight 1.5.

I would replace run_type conceptually with two values:

workload kind: page_list | category | linked
trusted source: webform | template | automated_tools | toolbar | ...

Then make the policy explicitly decide which dimensions affect cost.

5. HIGH/MEDIUM — “global” state is only global under an unstated single-pod assumption

The state defaults beneath sys_get_temp_dir(). That makes it process-shared within a machine/container filesystem, which is fine for multiple FastCGI workers in one pod.

But Toolforge supports multiple webservice replicas.  If Citation Bot is ever run with two replicas, each pod can have an independent /tmp/big-run.json, giving you two independent token buckets and two independent concurrency caps.

Thus BIG_RUN_MAX_TOTAL = 10 could effectively become 20, etc.

The repository contains no deployment configuration pinning this architectural requirement.

Either:

* explicitly document/enforce one web replica only, or
* move coordination to genuinely deployment-global storage such as Redis/database.

I would not put a high-frequency lock file on NFS merely to solve this; Toolforge’s storage guidance advises against frequently modified data on shared NFS. 

6. MEDIUM — trusted users are exempt from the exact capacity protection they most need

DEV_USERS are not merely allowed to exceed page limits; they also bypass the big-run gate altogether.  Category/linked entry points can grant those users a MAX_PAGES_OVERRIDE of 1,000,000.

“Trusted” answers the abuse question, but it does not answer the capacity question.

A trusted developer can accidentally consume a worker just as effectively as anyone else—and can launch much larger workloads than ordinary users.

I would consider:

* exempting developers from token/fair-use accounting;
* not exempting them from the global concurrency count;

with a separate explicit emergency/admin override if genuinely necessary.

7. MEDIUM — page count is a weak proxy for worker cost, and slow mode is ignored

The repository itself documents that slow mode performs extra external API work and can consume substantially more of the execution budget.

Yet:

5 fast pages = 5 slow pages

to the new token bucket.

Page complexity also varies enormously: one citation-heavy page can cost more than several simple pages.

The weighting scheme additionally assumes that template and automated_tools deserve one-third the per-page cost of a webform/category request, but I don’t see evidence in the PR that these ratios came from production measurements.

I would treat the initial constants as experimental and instrument:

* wall-clock duration;
* pages processed;
* slow/fast mode;
* source/workload type;
* external API call counts if practical.

Tune from observed worker-seconds rather than attribution labels.

8. MEDIUM — the cost cap stops distinguishing truly huge runs

A category-type large run reaches the 400-token maximum at roughly 178 pages. Everything beyond that costs the same.

So:

* 178 pages ≈ 400 tokens
* 1,000 pages = 400
* 10,000 pages = 400

Yet the amount of worker occupancy can be radically different.

The cap is understandable—otherwise a request could require more tokens than the bucket can ever contain—but it exposes another assumption: admission-time charging is sufficient for arbitrarily long work.

For genuinely large workloads, incremental charging/leases or chunking would model capacity much better.

9. MEDIUM — malformed state validation is inconsistent with the existing rate limiter

The pre-existing rate limiter carefully validates timestamps as finite/non-negative. The new state reader validates started_at only with is_numeric() and accepts any string as tier.

Consequences of corrupt state could include:

* a non-large tier still occupying total capacity but escaping the large count;
* a future/infinite started_at surviving pruning unexpectedly;
* nonsensical wait calculations.

Similarly, big_run_try_acquire() does not validate $page_count, $now, or $run_type as defensively as the sibling request_rate_limit_consume() implementation does.

This is not currently an HTTP exploit because callers constrain the values, but it is an unnecessary internal inconsistency.

10. MEDIUM — clock-change handling is only half implemented

The token bucket explicitly attempts to cope with the wall clock moving backward.

Active entries do not.

Staleness is based on:

now - started_at

so a significant forward clock correction can make live jobs instantly stale, while a backward correction can make entries linger considerably longer.

This strengthens the case for a heartbeat/lease model. If practical, use a monotonic measure for process-lifetime calculations.

11. MEDIUM — big_full means two different things

The code uses the same reason when:

* all 10 global bulk slots are occupied, or
* only the 4-large-run subpool is occupied.

A fifth large request can therefore display:

Citation Bot is currently at capacity with other big runs (4 in progress)

even though the global pool is only 40% occupied.

Worse, the wait calculation finds the oldest entry across all tiers, so a small job can determine the estimated wait for a large request even though that small job finishing does not free a large slot.

Use separate reasons such as total_full and large_full, and compute the blocking set accordingly.

12. MEDIUM/LOW — the PR contradicts the repository’s own constant-placement rule

The repository documentation says:

“Constants and definitions should be provided in constants.php.”

The PR adds all the BIG_RUN_* constants directly to RequestRateLimit.php, specifically because adding lines earlier would upset line-number-sensitive Progpilot suppressions.

That is a good example of one workaround generating another architectural inconsistency.

I would rather fix the fragile security-tool suppression arrangement than codify “don’t put lines above this function” as an architectural constraint.

13. LOW — overload logging can itself amplify overload

Every admission/deferral and lock-busy event calls error_log(). Under exactly the request storm this mechanism is supposed to mitigate, this can create a substantial logging storm.

The existing failure logger deliberately suppresses repeated messages per worker; the new normal-path logging doesn’t.

Rate-limit or sample the operational logs, and use counters/metrics if available.

⸻

The assumptions I would explicitly document

Before merging, I would want the design to state whether all of these are actually intended:

1. Citation Bot has exactly 24 usable dynamic-request workers.
2. Those workers all see one shared state file.
3. Production always has one Kubernetes replica.
4. ≤4-page requests are trusted not to be fragmented/parallelized abusively.
5. edit= attribution is sufficiently trustworthy to affect quota cost.
6. Page count correlates adequately with worker occupation.
7. Fast and slow processing cost approximately the same for admission purposes.
8. Developer runs need no concurrency protection.
9. A running bulk request normally finishes in well under five minutes.
10. A 120-second nominal duration is meaningful even though page/API processing can repeatedly reset execution limits.
11. Storage failures should sacrifice all bulk protection rather than merely degrade it.
12. It is acceptable that requests do substantial OAuth/API/list-discovery work before admission control.

Several of those are presently false, questionable, or at least unverified.

Revised merge recommendation

After this additional pass, I would strengthen my previous recommendation to do not merge yet.

The architecture can still be retained, but I would address these before calling the design reliable:

1. Verify and encode the actual production worker count/process model.
2. Remove requester-controlled quota discounts.
3. Fix the late category/linked admission problem.
4. Replace fixed started_at expiration with a renewable lease/heartbeat.
5. Make release reliable under short lock contention.
6. Prevent the existing per-user big-job check from wasting tokens.
7. Resolve the category source/type accounting contradiction.
8. Decide explicitly whether threshold splitting is acceptable; if not, add per-user small-request accounting.

At that point the gate would have a much stronger relationship between what it claims to guarantee and what it actually guarantees under production load.

redalert2fan:
All review findings are worked through at the current head. Blocker by blocker:

**Fixed (each with a red-first test):**
- Requester-controlled quota discounts — all page-list runs now bill at the webform rate (`big_run_charge_type`); only server-derived `category`/`webform_linked` keep their own weight.
- Per-user `big_jobs` check moved ahead of token spend, so rejected runs spend nothing.
- `big_run_release` retries briefly on lock contention instead of leaving ghost slots.
- Large-pool waits computed from large entries only (regression test proves the old-small-entry hijack is gone).
- Busy page served with 503 + `Retry-After` + `no-store`, mirroring the gadget limiter.
- Heartbeat leases (`last_seen_at`, renewed from the per-page loop) replacing `started_at` pruning; old-format entries fall back gracefully.
- Preflight pool check in `category.php`/`linked_pages.php` before remote discovery (advisory only — the full gate still decides admission, so no over-admission possible).

**Workers — resolved with primary evidence:** the live `~/lighttpd.conf` remaps `.php` → `.php8` at 1 proc × 24 `PHP_FCGI_CHILDREN` (the stock default is 2×2=4, which is where the confusion came from). So 24 stands; the process-model wording is corrected to php-cgi. Caps of 10/4 are coherent against 24 children.

**Verified intact:** progpilot whitelist (all pins ≤ line 95, gate edits start at 184+), CI green, all three entry points covered, cleanup refactor proven behavior-preserving by the suite.

**Maintainer decisions now recorded in `AGENTS.md`:** DEV bypass stays; threshold splitting accepted as a cooperative-shaping limitation; slow mode bills equally until production data says otherwise; gate logging stays at full volume until the 10/4/400 constants are validated against deferral rate and the `big_full`/`tokens` reason split in `error.log`.

GlazerMann:
I rebased it so that I can deploy it and play with it

redalert2fan:
Feel free to make changes as necessary ofcourse

GlazerMann:
[apply-pr5969-hardening-v5.7.php](https://github.com/user-attachments/files/32150306/apply-pr5969-hardening-v5.7.php)
[PR5969-revised-description-v5.7.md](https://github.com/user-attachments/files/32150311/PR5969-revised-description-v5.7.md)
@redalert2fan Please look over these possible changes suggest by AI.  The PHP file is a php run file patcher.

PR #5999, opened 2026-09-13
Add big-run gate so single requests take priority over bulk runs

GlazerMann:


GlazerMann:
# PR #5969 Hardening v5.9.1

## Purpose

`apply-pr5969-hardening-v5.9.1.php` is a standalone hardening patch for Citation Bot pull request #5969, **“Add big-run gate so single requests take priority over bulk runs.”**

The original PR introduced the basic big-run admission-control concept: web requests above four pages are gated so bulk work cannot consume every FastCGI worker. It added a shared concurrency limit, a large-run sub-limit, a token bucket, preflight checks for discovery-heavy entry points, and HTTP 503 responses for deferred work.

v5.9.1 keeps that goal and the reviewed admission economics, but substantially hardens the implementation around discovery, shared-state integrity, concurrency races, lease ownership, long-running network activity, per-user large-run locking, portability, observability, and deployment safety.

This document describes both:

1. what the resulting v5.9.1 code does; and
2. how it differs from the original PR #5969 implementation.

---

## Baseline and release identity

The patcher is deliberately pinned to the reviewed PR #5969 source tree at commit:

```text
d9cb2c47c8fff5ae2134ea969fb6a2474fa7cf13
```

It verifies the Git blob SHA-1 of every file it intends to modify before applying anything. If even one source file does not match the expected reviewed blob, the patcher refuses to proceed.

Patcher SHA-256:

```text
6720a9191d4dae6a66482c72bd62867ba1eb5992f96faf1b5c8262f8f480aff9
```

The patcher changes 19 reviewed files spanning runtime code, tests, documentation, and example configuration.

v5.9.1 is **standalone**. It is intended to be applied to a clean checkout of the pinned PR #5969 head, not layered on top of v5.8, v5.9, or another already-modified tree.

---

# 1. What the original PR #5969 does

The original PR introduces admission control for web runs larger than four pages.

Its principal behavior is:

- web runs of **1–4 pages** bypass the big-run gate;
- web runs above four pages enter a shared concurrency pool;
- the default total big-run limit is **10**;
- requests with at least **50 pages** are considered large, with a default large-run limit of **4**;
- a token bucket with default capacity **400** refills at **4.0 tokens/second**;
- admitted requests spend tokens based on page count, request type, and small/large classification;
- category and linked-page requests perform a preflight capacity check before expensive discovery;
- active entries contain a liveness timestamp and are heartbeated from the page-processing loop;
- deferred requests receive a 503 response and `Retry-After`;
- state is persisted in `big-run.json`.

This is a useful first implementation of capacity protection, but several parts of the original implementation are optimistic or best-effort in ways that can undermine the capacity guarantee under races, failures, malformed state, long discovery operations, or stale-process recovery.

---

# 2. What v5.9.1 does

## 2.1 Reserves capacity before remote discovery

The most important architectural change is that category and linked-page discovery no longer merely *check* capacity before the first upstream API request.

They now acquire a real, short-lived **probe lease before the first remote discovery call**.

The gate therefore has three shared lease phases:

```text
probe  ->  discovery  ->  running
```

A probe:

- uses a separate probe pool;
- defaults to a maximum of **4 concurrent probes**;
- consumes no tokens;
- does not consume the normal bulk-pool count;
- is obtained before the first category/linked-page discovery API request.

If discovery proves that the request is bulk work, the **same lease** is atomically promoted from `probe` to `discovery`. It does not drop the lease and race to acquire a second one.

If discovery proves that the request is small (four runnable pages or fewer), the temporary lease is released before normal page processing.

### Why this matters

The original preflight was side-effect free. Several processes could all observe “room available” and then simultaneously begin expensive discovery. The original check therefore reduced load in the common case but could not reserve worker capacity.

v5.9.1 converts that optimistic check into a real capacity reservation.

---

## 2.2 Preserves worker capacity explicitly

The reviewed defaults are:

```text
normal bulk leases:       10
discovery probe leases:    4
large leases:              4
token capacity:          400
token refill:            4.0/s
```

The live deployment discussed during review has 24 web workers.

With the defaults, this subsystem can account for at most:

```text
10 normal bulk/discovery/running workers
+4 probe workers
---------------------------------------
14 gate-controlled workers
```

That leaves approximately 10 workers available for smaller interactive work, gadget/API traffic, authentication, and other requests.

The documentation now makes the operational invariant explicit:

```text
WEB_WORKER_COUNT >
    CITATION_BOT_BIG_RUN_MAX_TOTAL
    + CITATION_BOT_BIG_RUN_MAX_DISCOVERY_PROBES
```

The gate is capacity reservation, not a scheduler. It does not reorder requests that have already reached FastCGI.

---

## 2.3 Uses an explicit lease state machine

Each persisted shared lease contains mandatory fields:

```text
started_at
tier
last_seen_at
phase
```

Valid phases are:

```text
probe
discovery
running
```

Valid tiers are:

```text
small
large
```

Transitions are checked under the shared lock.

Important transitions include:

```text
probe -> discovery
discovery/small -> discovery/large
discovery -> running
```

Unexpected or missing phases are not silently repaired.

This prevents several classes of race where a request could continue as though it still owned capacity after its state had been removed, replaced, or transitioned by another process.

---

## 2.4 Classifies large work during discovery

A request becomes “large” at **50 runnable pages**.

v5.9.1 does not wait until discovery is complete to apply the large-run subpool.

As soon as incremental discovery reaches the 50th runnable page:

1. the shared discovery lease is marked `tier=large`;
2. the four-large limit is checked atomically;
3. the existing per-user large-run lease is acquired;
4. further expensive discovery occurs only after those controls are in place.

This means discovery itself is subject to the large-run capacity limit.

### Original behavior

The original implementation primarily applied the large classification at final admission. A request could therefore perform substantial discovery before the final large-run accounting decision.

---

## 2.5 Bounds discovery work

The patch adds defense-in-depth limits to category and linked-page discovery.

Category enumeration is bounded to the accepted maximum plus one page, sufficient to know whether the request exceeded the limit.

Linked-page discovery is paginated rather than collecting an unbounded result before admission.

The category/linked discovery path also uses bounded signals such as:

- the fifth runnable page;
- a raw-candidate probe bound;
- a maximum of five continuation batches before a provisional lease must exist.

These limits protect against unusual API behavior, heavily filtered input, or continuation responses that otherwise could consume a worker indefinitely before admission.

---

## 2.6 Strengthens token accounting

The token economics remain fundamentally those introduced by PR #5969.

Defaults remain:

```text
capacity: 400
refill:   4.0 tokens/second
```

Tokens are now charged **exactly once**:

- at direct admission for a known page list; or
- during the atomic `discovery -> running` promotion.

A discovery probe spends no tokens.

Requester-controlled attribution is normalized inside the token-cost calculation. A user-controlled `?edit=` value cannot select a cheaper billing class.

Server-derived workload kinds such as category and linked-page work retain their intended billing classes.

Trusted operators can be token-exempt without being invisible to physical concurrency accounting.

---

## 2.7 Changes trusted-user behavior

The original PR documentation described `DEV_USERS` as bypassing the big-run gate entirely.

v5.9.1 narrows that exemption.

`DEV_USERS` retain their extended web page-count privileges and are **token-exempt**, but they still consume shared concurrency capacity.

Testing remains exempt from both token and concurrency accounting.

CLI bypasses the web gate only under the documented assumption that CLI work does not consume the same constrained web-worker pool.

This distinction protects physical worker capacity even when a trusted operator launches a large request.

---

# 3. Shared-state hardening

## 3.1 Separate permanent lock inode from replaceable state

### Original

The original code used the JSON state file itself as the lock target:

```text
big-run.json
```

The process opened the file, locked it, read it, truncated it, rewrote it, and flushed it.

### v5.9.1

The hardened design uses:

```text
big-run.lock   permanent flock() inode
big-run.json   replaceable state snapshot
```

Writers hold `big-run.lock`, create a same-directory temporary snapshot, fully write and flush it, and then atomically rename it over `big-run.json`.

This separates synchronization identity from state replacement.

It avoids changing or invalidating the lock inode when the JSON snapshot is atomically replaced.

---

## 3.2 Atomic state replacement

The hardened writer:

1. serializes the complete new state;
2. rejects a serialized result larger than the configured maximum;
3. creates a temporary file in the same directory;
4. writes all bytes, handling partial writes;
5. flushes the file;
6. uses `fsync()` when available as an additional best effort;
7. closes the temporary file;
8. atomically renames it over `big-run.json`.

The state file is not truncated in place.

This means a process should see either the previous complete snapshot or the new complete snapshot rather than an intentionally half-written state file.

---

## 3.3 Strict state directory and pathname handling

The state root must be a real, writable directory appropriate for the current process.

The hardening checks and protects against unsafe state paths, including symlinked lock/state paths.

Permissions are kept private:

```text
state directory: 0700
lock/state files: 0600
```

The state snapshot is capped at **64 KiB**.

This prevents an unexpectedly large or malformed state file from being treated as ordinary trusted internal state.

---

## 3.4 Strict schema validation

The original reader was intentionally permissive:

- missing or malformed top-level values could fall back toward a fresh bucket;
- malformed individual entries could be ignored;
- storage failures often failed open.

That is dangerous for a concurrency gate because malformed state can make active work disappear from accounting.

v5.9.1 treats an existing snapshot as valid only if the full structure is valid.

Required top-level values include:

```text
tokens
updated
entries
```

Every lease entry must contain valid:

```text
started_at
tier
last_seen_at
phase
```

One malformed persisted entry invalidates the whole snapshot instead of being silently discarded.

Known bulk/probe work then fails closed/retryable rather than proceeding unaccounted.

Small requests that never enter the gate are unaffected by state failure.

---

## 3.5 More precise invalid-state diagnostics

v5.9 introduced distinct state-failure reasons rather than combining several cases into one generic message.

The resulting code can distinguish:

```text
state_unreadable
state_empty
state_oversized
json_decode
top_level_schema
top_level_numeric
entry_schema
entry_value
```

Raw state contents are not written to logs.

This makes operational diagnosis significantly easier without exposing potentially sensitive internal data.

---

# 4. Lease ownership and stale-process behavior

## 4.1 Missing shared lease is now meaningful

The original heartbeat was best effort. A missing entry did not create a strong ownership-loss signal.

The hardened heartbeat returns an ownership status conceptually equivalent to:

```text
ok
retry
lost
```

Interpretation:

- `ok`: the request still owns its lease and renewal succeeded;
- `retry`: an infrastructure/lock/storage problem prevented a definitive answer;
- `lost`: a valid state snapshot exists, but this request's lease no longer exists.

`lost` is terminal.

A request whose stale lease was pruned cannot later wake up and continue processing outside the concurrency accounting.

---

## 4.2 Lease loss stops the request

When a running bulk request definitively loses its shared lease, the request stops rather than silently continuing.

Where HTTP output is still replaceable, the response is turned into a busy/error response.

The hardened HTTP behavior uses:

```text
503 Service Unavailable
Retry-After: ...
Cache-Control: no-store
```

This applies to admission failures and lease-loss situations where the response can still be controlled.

---

# 5. Long-running network liveness

## 5.1 Original heartbeat gap

Before v5.9, shared and per-user heartbeats could occur around page-loop activity and around network calls, but a single long or stalled `curl_exec()` could spend enough time inside libcurl that no PHP-level heartbeat occurred.

A healthy worker could therefore appear stale.

## 5.2 v5.9/v5.9.1 behavior

The common cURL path now heartbeats:

- before a transfer;
- during the transfer through libcurl's progress callback;
- after the transfer.

The existing cURL progress callback already enforced the response-size limit.

v5.9 extends that callback rather than replacing the safety behavior.

The callback therefore performs both:

```text
response-size enforcement
+
lease heartbeat work
```

The mandatory cURL progress/liveness options are reapplied before `curl_exec()` so caller-provided options cannot accidentally disable these invariants.

The progress heartbeat covers both:

- the shared big-run admission lease; and
- the per-user large-run lease when the request is in large-job mode.

---

# 6. Per-user >=50-page large-run lock hardening

The existing Citation Bot code already prevented one user from running multiple large jobs concurrently, but its implementation was based on pathname existence and timestamps in `/dev/shm`.

The original mechanism could race during stale takeover and shutdown cleanup.

v5.9.1 hardens this into a guarded ownership protocol.

## 6.1 Permanent per-user guard

A persistent `_guard` file serializes:

- acquisition;
- stale takeover;
- heartbeat;
- kill signaling;
- shutdown cleanup.

The guard prevents two contenders from simultaneously deciding that the same old lease is stale and both replacing it.

## 6.2 Inode ownership

The process keeps its lease file handle open.

Heartbeat and shutdown cleanup compare the held file's device/inode identity with the current pathname.

Therefore:

- an old resumed process cannot heartbeat a newer process's replacement lease;
- an old shutdown handler cannot unlink a newer request's lease;
- pathname reuse does not imply ownership.

Definitive ownership loss stops the old request.

## 6.3 Portable state location

The code prefers writable:

```text
/dev/shm
```

on systems where it exists.

If `/dev/shm` is unavailable or unwritable, it falls back to:

```text
sys_get_temp_dir()/citation-bot-big-jobs
```

This makes the large-job ownership mechanism usable on systems such as macOS while retaining the production Linux behavior.

---

# 7. HTTP output handling

Admission control can happen after some application code has begun constructing the response.

v5.9.1 uses a short-lived application-owned output buffer while the admission decision is still reversible.

If the request is deferred or loses ownership before the buffer is committed, the buffer can be discarded and replaced with a clean 503 response.

If admission succeeds, the buffer is flushed and normal progress streaming continues.

This is safer than sending partial successful HTML and later discovering that the request should have been rejected.

---

# 8. Original PR versus v5.9.1

| Area | Original PR #5969 | v5.9.1 |
|---|---|---|
| Small requests | <=4 pages bypass gate | Same |
| Normal bulk capacity | 10 | Same default |
| Large capacity | 4 for >=50 pages | Same default, enforced during discovery as well |
| Token bucket | 400 capacity, 4.0/s | Same default economics |
| Pre-discovery behavior | Side-effect-free preflight check | Real probe lease before first remote discovery call |
| Probe capacity | None | Separate default pool of 4 |
| Discovery accounting | Discovery can occur after optimistic check | `probe -> discovery -> running` reservation state machine |
| Large classification | Primarily final admission | At 50th runnable page during discovery |
| DEV_USERS | Gate bypass | Token-exempt but concurrency-accounted |
| Shared lock | `big-run.json` itself | Permanent `big-run.lock` inode |
| State write | Truncate/rewrite locked JSON | Complete temporary write + atomic rename |
| Invalid state | Permissive/fail-open tendencies | Strict whole-snapshot validation; bulk work fails closed/retryable |
| Malformed entry | Could be ignored | Invalidates snapshot |
| State size | No explicit small snapshot cap | 64 KiB cap |
| Symlink/path checks | Minimal | Private owned directory and symlink rejection |
| Shared heartbeat | Page-loop best effort | Ownership-aware status plus page loop and cURL hooks |
| Heartbeat during `curl_exec()` | No | Yes, progress callback |
| Lost shared lease | Could continue | Terminal stop |
| Per-user large lock | Path/timestamp based | Guarded stale takeover + inode ownership |
| `/dev/shm` dependency | Hard-coded | `/dev/shm` preferred, temp-dir fallback |
| HTTP defer output | Direct rendering | Admission buffer allows clean replacement with 503 |
| Discovery bounds | Basic entry-point logic | Bounded category and paginated linked discovery with additional continuation limits |
| Observability | `big_full`, `tokens`, `retry_later` etc. | Adds `probe_full`, `total_full`, `large_full`, `lease_lost`, detailed invalid-state reasons |
| Horizontal replicas | Implicit local-state limitation | Explicitly documented: local JSON + `flock()` is single-instance/shared-filesystem only |

---

# 9. Request flows

## 9.1 Direct known page list

For a direct webform page list:

```text
filter runnable pages
        |
        +-- <=4 pages --> process without shared big-run gate
        |
        +-- >4 pages
                |
                +-- per-user large checks as applicable
                |
                +-- shared admission
                        |
                        +-- capacity/token denied --> clean 503
                        |
                        +-- admitted --> running lease
                                            |
                                            +-- heartbeat during page work/cURL
                                            +-- release at shutdown
```

No discovery probe is needed because the page list is already known.

---

## 9.2 Category or linked-page request

```text
request begins
      |
      +-- acquire probe lease before first remote API call
      |
      +-- bounded discovery begins
              |
              +-- proves request remains <=4 pages
              |       |
              |       +-- release probe/discovery lease
              |       +-- process as small request
              |
              +-- proves request is bulk
                      |
                      +-- atomically promote probe -> discovery
                      |
                      +-- continue bounded discovery
                              |
                              +-- reaches 50 runnable pages
                              |       |
                              |       +-- mark discovery lease large
                              |       +-- acquire per-user large lease
                              |
                              +-- final page count known
                                      |
                                      +-- promote discovery -> running
                                      +-- charge tokens exactly once
                                      +-- process pages
```

At every transition the request must still own the expected lease and phase.

---

# 10. Failure semantics

The hardening intentionally distinguishes small interactive work from work that has entered the bulk gate.

## Small requests

A request that never enters the bulk/probe system is not made unavailable merely because the shared big-run state is broken.

## Known bulk/probe work

Once a request is known to require shared admission, failures in required gate state are conservative.

Examples include:

- invalid state schema;
- state too large;
- unreadable state;
- persistence failure;
- lock contention where ownership cannot be established;
- invalid phase transition;
- definitive shared lease loss.

These requests are deferred/stopped rather than permitted to run invisibly outside the accounting system.

This is a significant change from the original code's fail-open bias.

---

# 11. Configuration

Reviewed defaults are preserved:

```text
CITATION_BOT_BIG_RUN_MAX_TOTAL                 10
CITATION_BOT_BIG_RUN_MAX_DISCOVERY_PROBES       4
CITATION_BOT_BIG_RUN_MAX_LARGE                  4
CITATION_BOT_BIG_RUN_TOKEN_CAPACITY            400
CITATION_BOT_BIG_RUN_TOKEN_REFILL_PER_SECOND   4.0
```

Additional timing settings are configurable, including stale timeout, heartbeat interval, and pool retry time.

The heartbeat interval is bounded relative to the stale timeout so a normal lease has multiple opportunities to renew before it can legitimately be classified as stale.

The example environment file documents the available overrides without forcing non-default values.

---

# 12. Files changed

## Runtime

### `src/includes/RequestRateLimit.php`

Contains the hardened shared gate:

- configuration parsing;
- probe/discovery/running state machine;
- total/probe/large limits;
- token accounting;
- strict state parsing;
- atomic snapshot persistence;
- stale pruning;
- ownership-aware heartbeat handling;
- detailed structured reasons.

### `src/includes/WebTools.php`

Contains web-facing admission behavior:

- temporary admission output buffering;
- busy-page and HTTP response handling;
- probe acquisition/promotion;
- discovery-to-running promotion;
- lease release;
- integration with `edit_a_list_of_pages()`.

### `src/includes/WikipediaBot.php`

Adds bounded/paginated discovery primitives used by category and linked-page workflows.

### `src/category.php`

Uses the probe lease before the first category discovery request and promotes it as discovery proves the work is bulk/large.

### `src/linked_pages.php`

Uses the probe lease and paginated linked-page discovery.

### `src/process_page.php`

Passes the appropriate run classification through the direct page-list path.

### `src/includes/big_jobs.php`

Hardens the existing per-user >=50-page lock with:

- permanent guard inode;
- stale-takeover serialization;
- inode ownership checks;
- ownership-safe shutdown;
- heartbeat/kill serialization;
- portable state-directory fallback.

### `src/includes/bot_curl.php`

Preserves the existing response-size progress function while adding admission/large-job heartbeats before, during, and after transfers.

---

## Tests

The patch expands or updates tests in:

```text
tests/phpunit/includes/RequestRateLimitTest.php
tests/phpunit/includes/BotCurlTest.php
tests/phpunit/includes/WebToolsTest.php
tests/phpunit/includes/BigRunGatePageTest.php
tests/phpunit/includes/BigJobsCoverageTest.php
tests/phpunit/includes/bigJobTest.php
tests/phpunit/includes/WikipediaBotResponseCoverageTest.php
tests/run_tests.php
```

Coverage includes state validation, admission transitions, discovery bounds, ownership behavior, portability, cURL heartbeat integration, and HTTP gate behavior.

---

## Documentation/configuration

```text
docs/README.md
AGENTS.md
src/env.php.example
```

These files document the final operational model rather than the earlier optimistic preflight model.

---

# 13. What v5.9 specifically adds over v5.8

v5.9 is intentionally smaller than the earlier hardening releases.

It adds lifecycle assurance rather than changing admission economics.

Its major changes are:

1. shared and per-user heartbeats occur **inside long `curl_exec()` transfers** through the existing progress callback;
2. mandatory progress/liveness behavior is reapplied before execution;
3. malformed-state diagnostics become specific rather than generic;
4. regression tests cover malformed state and the heartbeat-aware cURL callback;
5. deployment documentation explicitly says that v5.8 -> v5.9 does not change the state schema or permanent lock inode.

There is no intentional change to:

- total capacity;
- probe capacity;
- large capacity;
- page threshold;
- large threshold;
- token capacity;
- token refill rate;
- token charging model.

---

# 14. What v5.9.1 changes over v5.9

v5.9.1 is a build-gate correction only.

It does not change runtime policy or state semantics.

It fixes:

- a stale PHPStan suppression in `src/category.php`;
- two redundant null-coalescing checks on the mandatory `phase` field in `RequestRateLimit.php`;
- final-newline handling for generated `bot_curl.php`, satisfying PHPCS while preserving the Progpilot-sensitive earlier line layout.

Therefore v5.9.1 should be understood as:

```text
v5.9 runtime behavior
+
clean static-analysis/style gates
```

not as another scheduling or capacity-policy revision.

---

# 15. Patcher safety behavior

The patcher itself is defensive.

## Exact source verification

Before making changes, it computes the Git blob SHA-1 for each target file and compares it with the reviewed expected value.

A mismatch aborts application.

This prevents the script from blindly performing text replacements against an unknown revision.

## Exact replacement assertions

Literal and regex edits expect exactly one intended match.

A missing or duplicated anchor causes an exception instead of guessing which occurrence to modify.

## `--check` mode

The patch can be tested for applicability without writing files:

```bash
php apply-pr5969-hardening-v5.9.1.php --check /path/to/citation-bot
```

or from the target checkout:

```bash
php /path/to/apply-pr5969-hardening-v5.9.1.php --check
```

## Atomic multi-file application

For actual application, the patcher:

1. stages complete replacement files beside each target;
2. records complete backups;
3. re-verifies the original blob immediately before each rename;
4. atomically renames staged files into place;
5. rolls already-replaced files back if a later replacement fails.

This also detects a concurrent edit between initial verification and commit.

---

# 16. Application

From a **clean checkout of PR #5969 at the pinned head**:

```bash
php /path/to/apply-pr5969-hardening-v5.9.1.php --check
php /path/to/apply-pr5969-hardening-v5.9.1.php
```

Do not apply the standalone patcher to a tree that already contains v5.8/v5.9 hardening.

For production gate-code replacement, quiesce the web service so old and new coordination protocols are not executing concurrently during the update.

v5.9/v5.9.1 uses the same state schema and permanent lock-inode protocol as v5.8, so a routine v5.8 -> v5.9.1 runtime upgrade does not itself require deleting `big-run.json`.

The historical migration from the original direct-`big-run.json` lock protocol to the permanent `big-run.lock` protocol **does** require a drained deployment.

---

# 17. Validation after application

The patcher prints the intended validation sequence:

```bash
php -l src/includes/RequestRateLimit.php
php -l src/includes/WebTools.php
php -l src/includes/WikipediaBot.php
php -l src/includes/big_jobs.php
php -l src/includes/bot_curl.php
php -l src/env.php.example

composer run test
composer run phpstan
composer run psalm
composer run psalm-taint
composer run phan
composer run phpcs
composer run progpilot

php tools/cs1_harness.php
php tools/cs1_harness.php --slow
```

These checks are important because the change touches concurrency/state handling as well as files monitored by static-analysis tooling.

`progpilot.json` contains line-sensitive suppressions for pre-existing code. The patch intentionally keeps the big-run subsystem below the sensitive region and avoids unnecessary line movement there.

---

# 18. Operational limits that remain

The hardened gate is intentionally not a general distributed scheduler.

It still has several deliberate boundaries:

- it reserves capacity; it does not reorder FastCGI requests;
- cooperative users can split work into multiple small requests;
- local JSON + `flock()` coordinates only one application instance or processes sharing the same filesystem and lock semantics;
- horizontal replicas need a genuinely shared transactional admission store;
- CLI bypass is safe only if CLI work does not consume the constrained web-worker envelope;
- the 10/4/4/400/4.0 defaults should be tuned from measured worker occupancy, latency, resource pressure, and deferral logs rather than treated as universal constants.

---

# 19. Summary

The original PR #5969 establishes the right high-level policy: keep small interactive requests responsive by limiting bulk work.

v5.9.1 retains that policy but changes the implementation from a largely optimistic, best-effort gate into an ownership-aware reservation system:

```text
original:
    preflight check
    -> discover
    -> final admission
    -> best-effort heartbeat
    -> in-place shared state

v5.9.1:
    reserved probe
    -> atomic discovery promotion
    -> large classification during discovery
    -> atomic running promotion / single token charge
    -> ownership-aware heartbeats during page work and cURL
    -> terminal handling of lost ownership
    -> strict, atomically replaced shared state
```

The practical result is a gate whose concurrency limits are much harder to bypass accidentally through races, stale processes, malformed state, long network transfers, or discovery work occurring before final admission.

GlazerMann:
5.9.2 pushed.

GlazerMann:
# PR #5969 Hardening v5.9.5

## Purpose

`apply-pr5969-hardening-v5.9.5.php` is a standalone hardening patch for Citation Bot pull request #5969, **“Add big-run gate so single requests take priority over bulk runs.”**

The original PR introduced the basic big-run admission-control model: larger web requests are constrained so bulk work cannot consume every PHP/FastCGI worker. The v5 hardening series preserves that policy while strengthening the implementation around discovery-time races, shared-state integrity, lease ownership, long-running network activity, per-user large-run ownership, portability, observability, deployment safety, and operator recovery.

v5.9.5 includes **two manual recovery interfaces** for strict shared-state validation:

1. **CLI recovery tool:** `tools/reset_big_run_state.php`
2. **Authenticated HTTP recovery endpoint:** `src/reset_big_run_state.php`

Both use the same recovery implementation in `src/includes/RequestRateLimit.php`.

There is intentionally **no automatic corruption recovery**. Invalid shared state keeps bulk admission fail-closed until an operator explicitly inspects or resets it.

---

## Release identity

The standalone patcher is pinned to the reviewed PR #5969 head:

```text
d9cb2c47c8fff5ae2134ea969fb6a2474fa7cf13
```

The patcher verifies the expected Git blob SHA-1 of existing files before modifying them and requires newly introduced files to be absent.

Patcher SHA-256:

```text
3963f2d370671fbeb3524d51b6697fc7d4cb9beef177605ff05aed45e78a10f6
```

v5.9.5 is intended to be applied to a **clean checkout of the pinned PR head**. It is not an incremental patch to apply over v5.9.4 or an earlier hardening release.

---

# 1. Original PR #5969 behavior

The original PR adds admission control for web runs above four pages.

Its main model is:

- requests of **1–4 pages** bypass the shared big-run gate;
- larger web requests enter a shared concurrency pool;
- total big-run concurrency defaults to **10**;
- runs of at least **50 pages** are considered large;
- large-run concurrency defaults to **4**;
- a token bucket defaults to **400 tokens**;
- tokens refill at **4.0 tokens/second**;
- token cost depends on page count, workload type, and small/large classification;
- category and linked-page entry points perform an early capacity preflight;
- active leases are heartbeated from page processing;
- deferred requests receive HTTP 503 and `Retry-After`;
- state is stored in `big-run.json`.

The policy is sound, but the original implementation has optimistic and fail-open behaviors around discovery, state parsing, stale leases, and storage that can weaken the capacity guarantee.

---

# 2. v5.9.5 resulting architecture

## 2.1 Explicit lease lifecycle

Shared leases use an explicit state machine:

```text
probe -> discovery -> running
```

Each persisted lease contains mandatory fields:

```text
started_at
tier
last_seen_at
phase
```

Valid phases:

```text
probe
discovery
running
```

Valid tiers:

```text
small
large
```

Transitions happen while holding the permanent shared lock and verify that the caller still owns the expected lease in the expected phase.

A valid state snapshot that no longer contains the caller's lease means ownership is lost; the caller does not silently continue outside shared accounting.

---

## 2.2 Real reservation before discovery

The original PR used a side-effect-free preflight check before category or linked-page discovery.

That could race:

```text
worker A checks -> room available
worker B checks -> room available
worker C checks -> room available
...
all start discovery
```

v5.9.5 instead acquires a real **probe lease before the first remote discovery request**.

Probe leases:

- use a separate pool;
- default to **4 concurrent probes**;
- spend no tokens;
- do not consume the normal 10-entry bulk pool;
- reserve discovery capacity before upstream API work begins.

If discovery proves the request is bulk work, the same lease is atomically promoted:

```text
probe -> discovery
```

There is no release-and-reacquire window.

If discovery proves the request remains small, the temporary lease is released.

---

## 2.3 Worker-capacity envelope

Reviewed defaults remain:

```text
normal bulk/discovery/running leases: 10
probe leases:                         4
large leases:                         4
token capacity:                     400
token refill:                       4.0/s
```

For the reviewed 24-worker deployment:

```text
10 normal gate-controlled workers
+4 probe workers
----------------
14 possible gate-controlled workers
```

This leaves approximately ten workers outside the bulk/probe envelope.

The operational invariant is:

```text
WEB_WORKER_COUNT >
    CITATION_BOT_BIG_RUN_MAX_TOTAL
    + CITATION_BOT_BIG_RUN_MAX_DISCOVERY_PROBES
```

The gate reserves capacity. It does not reorder requests already accepted by the web server/FastCGI layer.

---

## 2.4 Large-run enforcement during discovery

A request becomes large at **50 runnable pages**.

The hardened flow does not wait for all discovery to complete.

When discovery reaches the 50th runnable page:

1. the shared discovery lease is marked `tier=large`;
2. the shared four-large limit is checked atomically;
3. the per-user large-run lease is acquired;
4. additional expensive discovery proceeds only while those ownership controls remain valid.

This prevents the discovery stage itself from bypassing the large-run limit.

---

## 2.5 Bounded discovery

Category and linked-page discovery are bounded and incremental.

The hardened design uses:

- paginated/incremental discovery;
- the fifth runnable page as the small-to-bulk transition;
- the 50th runnable page as the large transition;
- bounded raw candidate probing;
- bounded continuation behavior before shared capacity must be reserved;
- maximum-page protection before accumulating unbounded upstream results.

This reduces unaccounted pre-admission work.

---

# 3. Token accounting

The admission economics introduced by the original PR are retained.

Defaults remain:

```text
capacity: 400
refill:   4.0 tokens/second
```

Tokens are charged exactly once:

- at direct admission when the page list is already known; or
- during the atomic `discovery -> running` promotion.

Probe leases spend no tokens.

Requester-controlled `?edit=` attribution cannot choose a cheaper billing class. Server-derived workload types retain their intended weights.

Trusted operators can remain token-exempt while still consuming physical concurrency slots.

---

# 4. Trusted-user behavior

The original PR treated `DEV_USERS` as bypassing the gate.

The hardened implementation narrows that exemption.

`DEV_USERS`:

- retain extended page-count privileges;
- may be exempt from token charging;
- still consume shared concurrency capacity.

Testing remains exempt.

CLI bypass assumes CLI jobs do not consume the same constrained web-worker envelope.

This distinction protects the physical worker pool even when a trusted operator launches a large request.

---

# 5. Shared-state storage

## 5.1 Permanent lock inode

Original design:

```text
big-run.json
    = state file
    + flock target
```

Hardened design:

```text
big-run.lock   permanent flock target
big-run.json   replaceable state snapshot
```

The lock inode remains stable while the state snapshot is atomically replaced.

---

## 5.2 Atomic state replacement

While holding `big-run.lock`, writers:

1. serialize a complete snapshot;
2. enforce the state-size limit;
3. create a temporary file in the same directory;
4. write all bytes;
5. flush it;
6. use `fsync()` when available as best effort;
7. close it;
8. atomically rename it over `big-run.json`.

The live JSON state is not truncated and rebuilt in place.

---

## 5.3 State-path protections

The state root must be a real writable directory appropriate for the process.

The hardening rejects unsafe conditions including symlinked shared state/lock paths.

Permissions are kept private:

```text
state directory: 0700
big-run.lock:    0600
big-run.json:    0600
recovery backup: 0600
```

The shared JSON snapshot is capped at **64 KiB**.

---

# 6. Strict whole-snapshot validation

Strict validation is deliberate.

An existing snapshot is accepted only when the entire structure is valid.

Required top-level data includes:

```text
tokens
updated
entries
```

Each persisted lease must contain valid:

```text
started_at
tier
last_seen_at
phase
```

Examples that invalidate the entire snapshot:

```text
malformed JSON
wrong top-level schema
invalid numeric values
invalid tier
invalid phase
negative timestamps
missing mandatory lease fields
oversized snapshot
```

A malformed lease is **not silently dropped**.

Known bulk/probe work fails closed/retryable because silently dropping an entry could forget a real active worker and admit more work than the configured limit permits.

---

## 6.1 Future timestamps

Future timestamps are handled separately from malformed data.

A structurally valid lease with future timestamps can be clamped to `now`.

The regression tests distinguish:

```text
malformed schema/value -> snapshot invalid -> fail closed
valid future timestamp -> clamp -> continue
```

This distinction was corrected in v5.9.2.

---

# 7. Invalid-state observability

The state parser reports specific reasons such as:

```text
state_unreadable
state_empty
state_oversized
json_decode
top_level_schema
top_level_numeric
entry_schema
entry_value
```

Raw snapshot contents are never logged.

When invalid state is detected, the runtime emits a prominent operator-facing message indicating that:

- the shared state is invalid;
- bulk admission remains fail-closed;
- automatic recovery will not occur;
- an operator should inspect or reset the state manually.

The operational invariant is:

> **There is intentionally no automatic corruption recovery. Manual recovery is available through the CLI tool or the password-protected HTTP endpoint.**

---

# 8. Operator recovery architecture

Both recovery front ends call the same functions in:

```text
src/includes/RequestRateLimit.php
```

The shared recovery backend consists primarily of:

```text
big_run_recovery_check()
big_run_recovery_reset()
```

Neither front end directly edits `big-run.json`.

This prevents the CLI and HTTP mechanisms from drifting into different state-repair implementations.

---

# 9. Recovery tool 1: CLI

## 9.1 Location

```text
tools/reset_big_run_state.php
```

The script is intentionally CLI-only.

If executed through a non-CLI SAPI, it refuses to operate.

---

## 9.2 Check mode

Command:

```bash
php tools/reset_big_run_state.php --check
```

The tool:

1. resolves the normal rate-limit state directory;
2. opens the permanent `big-run.lock`;
3. obtains the lock;
4. reads and validates the shared snapshot;
5. reports validity without changing state.

For valid state it reports:

```text
Big-run state: VALID
State path: ...
Persisted lease entries: ...
Token balance: ...
```

For invalid state it reports:

```text
Big-run state: INVALID
State path: ...
Reason: <validation reason>
Bulk admission remains fail-closed.
After draining/quiescing bulk workers, recover with:
  php tools/reset_big_run_state.php --reset
```

Exit behavior:

```text
0   valid state / help success
2   state check completed and state is invalid
64  invalid command-line usage
```

---

## 9.3 Reset mode

Command:

```bash
php tools/reset_big_run_state.php --reset
```

Before resetting, the CLI prints:

```text
WARNING: reset forgets every shared big-run lease.
Drain/quiesce bulk workers first.
```

The reset then uses the common recovery backend.

On success it reports:

- state path;
- preserved backup path, if a previous snapshot existed;
- confirmation that the fresh state has:
  - zero leases;
  - a full token bucket.

Reset failure returns exit code `3`.

---

# 10. Recovery tool 2: authenticated HTTP endpoint

## 10.1 Location

```text
src/reset_big_run_state.php
```

This is a **manual privileged maintenance endpoint**, not automatic self-healing.

It is deliberately modeled after:

```text
src/gitpull.php
```

---

## 10.2 Authentication

The endpoint uses the existing:

```text
DEPLOY_PASSWORD
```

The same secret therefore protects both privileged maintenance functions unless deployment configuration chooses otherwise in the future.

Authentication uses:

```php
hash_equals($deployPassword, $password_in)
```

for constant-time password comparison.

If `DEPLOY_PASSWORD` is missing or blank, the recovery endpoint refuses service with HTTP 503.

---

## 10.3 Interactive browser use

A browser GET to:

```text
/reset_big_run_state.php
```

shows two forms:

```text
Check state
Reset state
```

Both require the deployment password.

The reset form additionally requires a checkbox confirming:

> I have drained or quiesced bulk workers and understand that reset forgets all shared leases.

GET never performs a check or reset operation itself.

---

## 10.4 Automation

Like `gitpull.php`, automation can send the password in:

```text
X-Deploy-Token
```

Example check:

```bash
curl -X POST   -H "X-Deploy-Token: $DEPLOY_PASSWORD"   -d 'action=check'   https://HOST/reset_big_run_state.php
```

Example reset:

```bash
curl -X POST   -H "X-Deploy-Token: $DEPLOY_PASSWORD"   -d 'action=reset&confirm_reset=yes'   https://HOST/reset_big_run_state.php
```

A reset is not accepted merely because the caller supplied a valid password. The explicit confirmation value is also required:

```text
confirm_reset=yes
```

---

## 10.5 Query-string protection

Credentials and recovery actions are never accepted from the query string.

If GET contains parameters, the endpoint redirects to the clean endpoint URL.

The implementation does not consume:

```text
?password=...
?action=reset
```

This avoids credential leakage through:

- browser history;
- proxy/access logs;
- referrers;
- copied URLs.

---

## 10.6 Request-method handling

Supported methods:

```text
GET   display the form only
POST  authenticated check/reset action
```

Other methods receive:

```text
405 Method Not Allowed
Allow: GET, POST
```

---

## 10.7 HTTP hardening

The endpoint mirrors the privileged `gitpull.php` protection pattern.

It loads:

```text
env.php
includes/PublicConfig.php
includes/RequestRateLimit.php
```

and applies public host/origin validation.

Security headers include:

```text
Cache-Control: no-store
Referrer-Policy: no-referrer
X-Content-Type-Options: nosniff
Content-Security-Policy:
    default-src 'none';
    form-action 'self';
    base-uri 'none';
    frame-ancestors 'none'
```

Response text is escaped with `htmlspecialchars()`.

The dependency set is deliberately small so recovery remains usable even if unrelated application code is unhealthy.

---

## 10.8 HTTP status behavior

Representative results include:

```text
200  form displayed
200  valid-state check
200  successful reset
400  missing/invalid password submission
400  missing/invalid action
400  reset without explicit confirmation
403  incorrect password
405  unsupported method
409  state check found invalid state
409  reset could not obtain permanent lock
503  DEPLOY_PASSWORD unavailable
503  other reset infrastructure failure
```

An invalid-state check returns a message stating that bulk admission remains fail-closed and that **no automatic recovery was attempted**.

---

# 11. Common recovery backend

The two front ends share all important recovery semantics.

## 11.1 Check semantics

`big_run_recovery_check()`:

- resolves the configured state location;
- opens the permanent lock safely;
- obtains `big-run.lock`;
- validates the snapshot with the production parser;
- returns:
  - valid/invalid;
  - validation reason;
  - state path;
  - lease count when valid;
  - token balance when valid.

It does not modify the snapshot.

---

## 11.2 Reset semantics

`big_run_recovery_reset()` is explicitly operator-requested.

The caller should drain/quiesce bulk workers first.

While holding `big-run.lock`, it:

1. rejects a symlinked state path;
2. requires any existing state path to be a regular file;
3. chooses a unique same-directory recovery backup name;
4. renames the existing snapshot to that backup;
5. sets the backup permission to `0600`;
6. creates a fresh state atomically;
7. initializes:
   - token balance to full capacity;
   - lease list to empty;
   - update time to the current resolved time;
8. clears the remembered invalid-state reason;
9. records an `operator_state_reset` log entry.

---

## 11.3 Backup naming

A preserved state snapshot uses a name of the form:

```text
big-run.json.recovery-YYYYMMDDTHHMMSSZ-PID
```

If that name already exists, a numeric suffix is added.

Up to 100 candidate names are tried.

These backups are forensic/operator artifacts and should not be treated as active state.

---

## 11.4 Write-failure rollback

If the old snapshot was successfully preserved but creation of the fresh state fails, the reset routine attempts to rename the backup back to `big-run.json`.

Therefore the expected sequence is:

```text
old state
   |
   +-- rename to backup
   |
   +-- attempt fresh state
           |
           +-- success -> keep backup
           |
           +-- failure -> restore old state if possible
```

If both the fresh write and rollback fail, the runtime emits a high-priority failure indicating that manual recovery is required.

This is safer than deleting malformed state first and only then attempting to construct replacement state.

---

# 12. Why workers should be drained first

A reset deliberately forgets every shared lease.

If active bulk workers continue running while reset occurs, the fresh snapshot no longer contains their old lease IDs.

The existing ownership-loss logic means those workers should subsequently detect that their lease is gone and stop rather than silently recreating ownership.

Even so, the safest operational procedure is:

```text
1. drain or quiesce bulk workers
2. inspect state
3. preserve diagnostic evidence
4. reset state
5. restart/resume traffic
```

The browser reset form forces an explicit acknowledgement of this requirement.

The CLI prints the same warning, but—as appropriate for an operator shell tool—it does not implement an interactive password/confirmation prompt.

---

# 13. No automatic corruption recovery

The presence of two reset tools does **not** change the core fail-closed design.

The runtime does not:

- silently drop malformed persisted leases;
- rename an invalid snapshot automatically;
- reset the token bucket automatically;
- automatically clear all leases after a timeout;
- invoke the HTTP endpoint itself;
- invoke the CLI recovery tool itself.

Recovery is always an explicit operator decision.

This preserves correctness: a malformed entry could represent active work whose structure is no longer understood, so silently erasing it could over-admit bulk work.

---

# 14. Shared lease ownership and stale-process behavior

Heartbeat operations distinguish approximately:

```text
ok
retry
lost
```

Meaning:

- `ok`: the process still owns its lease;
- `retry`: infrastructure prevented a definitive answer;
- `lost`: the snapshot is valid but the process's lease is absent.

`lost` is terminal.

A worker whose stale lease was pruned or whose lease disappeared during an operator reset cannot later resume untracked.

---

# 15. Heartbeats during long network transfers

The common cURL path heartbeats:

```text
before curl_exec()
during curl_exec() through the progress callback
after curl_exec()
```

The existing progress callback's response-size enforcement is preserved.

The hardened callback therefore enforces both:

```text
maximum response size
+
shared/per-user lease liveness
```

Mandatory cURL progress options are reapplied before transfer execution so callers cannot accidentally disable those invariants.

---

# 16. Per-user >=50-page ownership hardening

The pre-existing Citation Bot large-job mechanism prevented one user from launching multiple large runs, but relied heavily on pathname existence/timestamps.

The hardening introduces:

- a permanent per-user `_guard`;
- serialized acquisition/stale takeover;
- held lease file handles;
- inode/device ownership checks;
- ownership-safe heartbeat;
- ownership-safe shutdown cleanup;
- ownership-safe kill signaling.

An old process cannot unlink or heartbeat a newer process's replacement lease merely because the pathname is the same.

---

## 16.1 Portable storage

The per-user large-job mechanism prefers writable:

```text
/dev/shm
```

but can fall back to:

```text
sys_get_temp_dir()/citation-bot-big-jobs
```

This preserves Linux production behavior while improving portability.

---

# 17. HTTP admission behavior

Deferred bulk work uses:

```text
503 Service Unavailable
Retry-After: ...
Cache-Control: no-store
```

The hardening uses short-lived application-owned output buffering while an admission decision is still reversible.

If the request is rejected or loses ownership before output is committed, the partial buffer can be discarded and replaced with a clean error response.

---

# 18. Original PR versus v5.9.5

| Area | Original PR #5969 | v5.9.5 |
|---|---|---|
| Small requests | <=4 pages bypass gate | Same |
| Normal bulk capacity | 10 | Same default |
| Large capacity | 4 for >=50 pages | Same default, enforced during discovery |
| Token bucket | 400 / 4.0 per second | Same economics |
| Pre-discovery protection | Optimistic preflight | Reserved probe before first remote call |
| Probe capacity | None | Separate pool, default 4 |
| Lease phases | Primarily active entry | `probe -> discovery -> running` |
| Large classification | Primarily final admission | At 50th runnable page during discovery |
| DEV_USERS | Full gate bypass | Token-exempt but concurrency-accounted |
| State lock | `big-run.json` inode | Permanent `big-run.lock` |
| State writing | Truncate/rewrite | Atomic replacement |
| Malformed entry | Could be discarded/permissive | Entire snapshot invalid |
| State size | No explicit small cap | 64 KiB |
| Symlink/path defense | Minimal | Explicit safety checks |
| Lease-loss semantics | Best effort | Definitive loss is terminal |
| Heartbeat during cURL | No | Yes |
| Per-user large lock | Path/timestamp oriented | Guard + inode ownership |
| `/dev/shm` dependency | Hard-coded | Portable fallback |
| Invalid-state recovery | None | CLI + authenticated HTTP manual recovery |
| Automatic recovery | N/A | Explicitly disabled |
| Reset backup | None | Timestamped `.recovery-*` forensic backup |
| HTTP reset authentication | None | `DEPLOY_PASSWORD`, `hash_equals`, POST-only action |
| HTTP automation | None | `X-Deploy-Token` |
| Reset confirmation | None | Required `confirm_reset=yes` over HTTP |
| Horizontal replicas | Local-state limitation | Explicitly documented limitation |

---

# 19. The two reset tools compared

| Property | CLI tool | HTTP endpoint |
|---|---|---|
| Path | `tools/reset_big_run_state.php` | `src/reset_big_run_state.php` |
| Intended use | Shell/on-call maintenance | Browser or authenticated automation |
| Check support | `--check` | `action=check` |
| Reset support | `--reset` | `action=reset` |
| Password required | No; relies on shell/operator access | Yes, `DEPLOY_PASSWORD` |
| Automation credential | N/A | `X-Deploy-Token` |
| Explicit reset acknowledgement | Warning printed | Required `confirm_reset=yes` |
| GET can reset | N/A | No |
| Query-string credentials/actions | N/A | Rejected/not consumed |
| Backend | `big_run_recovery_*()` | Same |
| Permanent lock | Yes | Yes |
| Backup preservation | Yes | Yes |
| Atomic fresh state | Yes | Yes |
| Automatic recovery | No | No |

The CLI tool remains the preferred recovery mechanism when shell access is available because it minimizes exposed surface area.

The HTTP endpoint exists for controlled operational convenience and mirrors the existing deployment endpoint's authentication boundary.

---

# 20. Recovery runbook

## Preferred CLI procedure

### Step 1: drain/quiesce bulk work

Stop new bulk admission and allow current bulk jobs to drain where practical.

### Step 2: inspect

```bash
php tools/reset_big_run_state.php --check
```

Record the reported reason.

### Step 3: preserve external diagnostics if required

The reset itself preserves the existing regular snapshot, but operators may also want relevant application logs or filesystem metadata before reset.

### Step 4: reset

```bash
php tools/reset_big_run_state.php --reset
```

### Step 5: verify

```bash
php tools/reset_big_run_state.php --check
```

Expected fresh state:

```text
VALID
0 leases
full token balance
```

### Step 6: resume service

Resume bulk admission and observe gate logs.

---

## Authenticated HTTP procedure

### Interactive browser

Visit:

```text
https://HOST/reset_big_run_state.php
```

Use **Check state** first.

If reset is required:

1. drain/quiesce bulk workers;
2. enter the deployment password;
3. select the reset confirmation checkbox;
4. submit **Reset shared state**;
5. use **Check state** again.

### Automation

Check:

```bash
curl -X POST   -H "X-Deploy-Token: $DEPLOY_PASSWORD"   -d 'action=check'   https://HOST/reset_big_run_state.php
```

Reset:

```bash
curl -X POST   -H "X-Deploy-Token: $DEPLOY_PASSWORD"   -d 'action=reset&confirm_reset=yes'   https://HOST/reset_big_run_state.php
```

The secret should be supplied through the header rather than a URL.

---

# 21. Files involved in recovery

## `src/includes/RequestRateLimit.php`

Contains the common recovery backend:

```text
big_run_recovery_check()
big_run_recovery_backup_path()
big_run_recovery_reset()
```

It also contains the normal gate parser, state writer, lock handling, and invalid-state diagnostics.

---

## `tools/reset_big_run_state.php`

CLI front end.

Responsibilities:

- CLI-only enforcement;
- argument validation;
- human-readable check output;
- operator reset warning;
- exit codes;
- calling the common recovery backend.

It does not implement separate state manipulation.

---

## `src/reset_big_run_state.php`

Authenticated HTTP front end.

Responsibilities:

- `DEPLOY_PASSWORD` authentication;
- browser form;
- `X-Deploy-Token` automation;
- GET/POST restrictions;
- explicit reset acknowledgement;
- security headers;
- host/origin validation;
- escaped HTML output;
- HTTP status mapping;
- calling the common recovery backend.

It does not implement separate state manipulation.

---

## `tests/phpunit/includes/BigRunRecoveryEndpointTest.php`

Regression coverage verifies that the HTTP endpoint retains the intended privileged boundary, including:

- `env.php`/`PublicConfig.php`;
- host/origin validation;
- `DEPLOY_PASSWORD`;
- `X-Deploy-Token`;
- `hash_equals()`;
- GET vs POST restrictions;
- no-store/no-referrer/nosniff/CSP headers;
- no query-string password/action path;
- authentication before recovery;
- explicit reset confirmation before reset.

---

# 22. Version-specific history

## v5.8

Established the hardened lease/state architecture, including:

- probe pool;
- early reservation;
- atomic phase promotion;
- permanent lock inode;
- strict state validation;
- improved per-user ownership behavior.

---

## v5.9

Lifecycle assurance:

- heartbeat before, during, and after cURL transfers;
- progress callback preserves response-size enforcement;
- more precise invalid-state diagnostics;
- lifecycle regression tests;
- deployment documentation.

No admission economics changed.

---

## v5.9.1

Build-gate correction:

- removed a stale PHPStan suppression;
- removed redundant `?? null` checks on mandatory `phase`;
- restored the required final newline in generated `bot_curl.php`.

No runtime policy changed.

---

## v5.9.2

Regression-test consistency correction:

- replaced the obsolete expectation that malformed entries are individually dropped;
- explicitly tests that malformed schema/value invalidates the whole snapshot;
- separately tests that valid future timestamps are clamped.

No runtime policy changed.

---

## v5.9.3

Operational recovery tooling:

- adds `big_run_recovery_check()`;
- adds `big_run_recovery_reset()`;
- adds `tools/reset_big_run_state.php`;
- preserves malformed/old state as `.recovery-*`;
- adds locked, atomic reset behavior;
- adds prominent invalid-state operator guidance.

At this point recovery is CLI-only.

---

## v5.9.4

Authenticated HTTP operational recovery:

- adds `src/reset_big_run_state.php`;
- mirrors `gitpull.php`'s `DEPLOY_PASSWORD` / `X-Deploy-Token` model;
- uses `hash_equals()`;
- GET shows forms only;
- POST performs actions;
- query-string credentials/actions are not accepted;
- reset requires explicit confirmation;
- both HTTP actions call the same v5.9.3 recovery backend;
- automatic recovery remains disabled.

---

## v5.9.5

PHPCS build-gate correction:

- replaces the new regression test's plain `strpos()` calls with `mb_strpos()`;
- fixes the `big_run_log_state_invalid()` PHPDoc so `$reason` and `$fields` match the actual function parameters.

v5.9.5 changes **no admission, reset, authentication, state, token, or lease behavior** from v5.9.4.

It is the v5.9.4 behavior with the reported PHPCS issues corrected.

---

# 23. Patcher safety

The standalone patcher uses exact source verification.

It:

1. checks the target checkout;
2. computes Git blob IDs;
3. verifies each existing target against the reviewed expected blob;
4. requires new files to be absent;
5. performs exact literal/regex substitutions;
6. stages complete replacement files;
7. re-verifies originals before commit;
8. atomically renames replacements into place;
9. rolls already-applied files back if a later commit fails.

The new CLI tool is created with executable permissions.

---

## Check-only application

Before applying:

```bash
php /path/to/apply-pr5969-hardening-v5.9.5.php --check
```

Then apply:

```bash
php /path/to/apply-pr5969-hardening-v5.9.5.php
```

Use a clean checkout at the pinned PR head.

---

# 24. Deployment notes

The historical migration from the original PR's direct `big-run.json` locking model to the hardened permanent `big-run.lock` protocol should be performed with the web/bulk workload drained.

Once already operating on the v5.8+ permanent-lock/state protocol, the v5.9 lifecycle/recovery releases do not intentionally introduce another shared-state schema migration.

For a production code replacement:

1. quiesce relevant web/bulk activity;
2. deploy the new files;
3. run validation;
4. restart/resume service;
5. verify the recovery endpoint is protected by the intended `DEPLOY_PASSWORD`;
6. perform a non-destructive recovery check if desired.

Do not delete `big-run.lock` as part of routine recovery.

---

# 25. Validation

After application, run the normal project validation suite:

```bash
php -l src/includes/RequestRateLimit.php
php -l src/includes/WebTools.php
php -l src/includes/WikipediaBot.php
php -l src/includes/big_jobs.php
php -l src/includes/bot_curl.php
php -l src/reset_big_run_state.php
php -l tools/reset_big_run_state.php
php -l src/env.php.example

composer run test
composer run phpstan
composer run psalm
composer run psalm-taint
composer run phan
composer run phpcs
composer run progpilot

php tools/cs1_harness.php
php tools/cs1_harness.php --slow
```

The recovery path should also be functionally exercised in a safe test/staging state:

```bash
php tools/reset_big_run_state.php --check
```

For HTTP testing, verify at minimum:

```text
GET form                         -> 200
incorrect password              -> 403
authenticated check             -> 200 or 409 depending on state
reset without confirmation      -> 400
unsupported method              -> 405
missing DEPLOY_PASSWORD         -> 503
authenticated confirmed reset   -> 200 in an approved test state
```

Do not perform a production reset merely as a smoke test.

---

# 26. Remaining architectural limitations

The hardened gate is still intentionally not a distributed scheduler.

Important limitations remain:

- it reserves capacity but does not reorder already-accepted FastCGI requests;
- users can split work into multiple small requests;
- JSON + `flock()` assumes processes share the same filesystem and lock semantics;
- independently hosted replicas require a genuinely shared transactional admission store;
- CLI bypass is safe only if CLI work does not consume the constrained web-worker envelope;
- capacity/token defaults should be tuned from measured production behavior;
- the authenticated HTTP recovery endpoint is an operational convenience and increases exposed maintenance surface compared with CLI-only recovery, so access to `DEPLOY_PASSWORD` must remain tightly controlled.

---

# 27. Summary

The original PR provides the high-level admission policy:

```text
protect small requests by limiting bulk work
```

v5.9.5 keeps that policy but hardens the mechanism into an ownership-aware reservation system:

```text
reserved probe
    -> bounded discovery
    -> atomic discovery promotion
    -> large classification during discovery
    -> atomic running promotion / one token charge
    -> ownership-aware heartbeats
    -> terminal lease-loss handling
    -> strict atomic shared state
```

Strict whole-snapshot validation intentionally creates a fail-closed operational condition when persisted state is malformed.

v5.9.5 addresses that operational consequence without weakening correctness:

```text
invalid state
    -> bulk work remains fail-closed
    -> no automatic repair
    -> operator inspects state
         via CLI
         or DEPLOY_PASSWORD-protected HTTP
    -> operator drains/quiesces bulk workers
    -> explicit reset
    -> old snapshot preserved
    -> fresh empty/full-token state installed atomically
```

Both reset interfaces use the same permanent-lock recovery implementation, so the HTTP endpoint adds an authenticated operational surface rather than a second or weaker reset mechanism.

